PR #27, opened 2021-09-16
Adds user's current allocation in navbar

### LINKED ISSUE #5: Add current allocation below the name of the user
We should add the current allocation below the name of the user in the navbar.

Motivation: Highlight the information in order to let users notify us about inconsistencies.

image

LucasHFS:
Added the current allocation below the name of the user in the navbar.

Closes #5

image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -43,8 +43,13 @@ class ApplicationController < ActionController::Base
     @current_company ||= current_user.company
   end
 
+  def current_user
+    UserDecorator.decorate(super) unless super.nil?
+  end
+
   protected
-    def configure_permitted_parameters
-      devise_parameter_sanitizer.permit(:sign_in, keys: [:email, :password, :otp_attempt])
-    end
+
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:sign_in, keys: [:email, :password, :otp_attempt])
+  end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -37,7 +37,9 @@ def after_sign_in_path_for(user)
     end
   end
 
-  private
+  def current_user
```
wenderjean, line 40: Would we have issues moving this method under the `protected` session?
Can we create a spec to ensure the object returned will be always a decorated user?

adbatista (changes requested):
missing specs

spec/decorators/user_decorator_spec.rb
```diff
@@ -3,21 +3,37 @@
...
+    context 'when the current user is an admin and user is allocated and' do
       it 'returns project link' do
-        link = "<a href=\"http://localhost:3000/admin/projects/#{allocation.project_id}\">#{CGI.escape_html(allocation.project.name)}</a>"
-
+        link = "<a href=\"/admin/projects/#{allocation.project_id}\">#{CGI.escape_html(allocation.project.name)}</a>"
         expect(user.current_allocation).to eq(link)
       end
     end
 
+    context 'when the current user is not an admin and user is allocated and' do
```
wenderjean, line 25: There is an ending `and` here :nail_care:

wenderjean (approved):
LGTM

adbatista (approved):

Issue #11, opened 2021-08-30
Apply a better design in the profile change page


wenderjean:
We should fix the design of the profile change page in order to make it similar to the profile index.

`/user/edit`

![image](https://user-images.githubusercontent.com/941776/131350964-6784ff61-37c0-466e-8099-541f70626543.png)

Issue #10, opened 2021-08-30
Apply a better design in the password change page


wenderjean:
We should fix the design of the password change page in order to make it similar to the profile one.

`users/account/password/edit`

![image](https://user-images.githubusercontent.com/941776/131350596-dabd84e1-1176-4eb3-9832-fd39c5ad4041.png)

PR #30, opened 2021-09-21
Improve single page forms

marmitoTH:
This PR addresses issues #10 and #11 

The idea behind this improvement is to make all "single page forms" share a single stylesheet, this is why I'm fixing both issues at the same time. The user update form and the password change now are centralized and have better spacing between elements. I also noticed that the user update form was not using the `form-control` class on its HTML fields, making the styling a bit different from other forms. 

The Form Container was also designed to be responsive, but since the rest of the page is not, there's not much I can do right now.

<details>
<summary>Screenshots</summary>

localhost_5000_user_edit
localhost_5000_users_account_password_edit
</details>

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -20,6 +20,7 @@
 @import 'deactivate_two_factor';
 @import 'pagination';
 @import 'show_user';
+@import 'form_container';
 
 $gray: #ededed;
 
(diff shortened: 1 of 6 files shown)


adbatista:
can you fix the cancel button stylesheet to remove the blue color of link?

marmitoTH:
> can you fix the cancel button stylesheet to remove the blue color of link?

Done. I removed the !important property from it, all other anchors seem to be styled properly with the change.

Issue #9, opened 2021-08-30
Disable "Save" button when there is no project selected


wenderjean:
On the dashboard page, we should not enable the "Save" button unless a project is selected.

![image](https://user-images.githubusercontent.com/941776/131349867-e570ea81-3003-420b-bc62-f191e28f9930.png)

PR #33, opened 2021-09-22
Disable save button when no project is selected

marmitoTH:
This PR addresses issue #9 

The value of the project selector is now being stored in a state in order to control the disabled state of the save button.
The dashboard tests have been updated to select a project before trying to save the form.
A new test has been created to validate if the button is really disabled when no project is selected.


adbatista:
can you post a gif or a video of the fix?

PR #35, opened 2021-09-22
Add "Administrative" item to admin users on navbar

### LINKED ISSUE #25: Add "Administrate" item to the navbar
An user may be both an engineer and an administrator in Punch, on those cases we want to add a new link to the navbar to avoid type "/admin" whenever we want to navigate to the panel.

ramos-ph:
Currently, to access the administrative dashboard, the user must type "/admin" at the end of the URL. This PR fixes this by adding an "Administrative" link on the header of the page.

If the current user is an admin:
image

Otherwise:
image

Closes #25

### DIFF
--- a/app/views/refills/_navigation.html.erb
+++ b/app/views/refills/_navigation.html.erb
@@ -21,6 +21,12 @@
           <%= link_to repositories_path do %>
             <%= t('.open_source') %>
           <%- end %>
+
+          <% if current_user.is_admin? %>
+            <%= link_to "/admin" do %>
+              <%= t('.administrative') %>
+            <%- end %>
+          <%- end %>
         </div>
 
         <div class="navbar-nav ml-auto nav-link">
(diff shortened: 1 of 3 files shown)


app/views/refills/_navigation.html.erb
```diff
@@ -21,6 +21,12 @@
           <%= link_to repositories_path do %>
             <%= t('.open_source') %>
           <%- end %>
+
+          <% if current_user.is_admin? %>
+            <%= link_to "/admin" do %>
```
ramos-ph, line 26: I haven't found `admin_path` related in the routes file. Please let me know if there is a better way of doing it.

geeksilva97: Use `admin_root_path`.

wenderjean (approved):

app/views/refills/_navigation.html.erb
adbatista: @ramos-ph fix this please so we can merge this PR

adbatista (changes requested):

adbatista (approved):

Issue #17, opened 2021-08-31
Simplify Punches table


wenderjean:
Is blocked by: https://github.com/Codeminer42/Punchclock/issues/16

We may want to simplify the Punches table by showing a single line per day, in our database an entire labor day is represented by two rows in the punches table and what we do is show those rows separately. Turns out that it causes too much noise on the page, a better UX would be to show a single row per day using the beginning hour for the day and the last hour regardless it was in the morning or afternoon.

PR #38, opened 2021-09-23
Simplify punches table

### LINKED ISSUE #16: Fix columns in our Punches table
The Punches table seems to have more columns in the body than in the header.

image

geeksilva97:
Fixes #16 and #17

This PR merges the two parts of a Punch and displays as one. The punch table is now as shown below.

print

### DIFF
--- a/app/controllers/punches_controller.rb
+++ b/app/controllers/punches_controller.rb
@@ -11,6 +11,7 @@ class PunchesController < ApplicationController
 
     @search.sorts = 'from desc' if @search.sorts.empty?
     @punches = PunchesPaginationDecorator.new(params, @search.result)
+    @grouped_punches = @punches.group_by { |punch| punch.when }.values
   end
 
   def new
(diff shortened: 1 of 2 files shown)


adbatista:
can you add a screenshot?

geeksilva97:
I think this solves the ISSUES mentioned, but some questions arise.

1. What about delete a Punch? There is only one button to delete that is addresses to only one piece of merged Punch (same for showing and editing)

geeksilva97:
> I think this solves the ISSUES mentioned, but some questions arise.
> 
> * What about delete a Punch? There is only one button to delete that is addresses to only one piece of merged Punch (same for showing and editing)

@wenderjean @adbatista what do you think?

wenderjean:
@geeksilva97 The ideal scenario would be a clickable row that when clicked would show all related punches for the day below it but It would be complex to implement even because we don't have tooling that favors complex front-end interactions. In this case, we can remove the "trash" button in this ticket and create another one to create a page to detail a given punch and there we would be able to remove each line.

What do you think @adbatista?

adbatista:
I don't think it is worth it, punches are independent the dashboard is just a convenience

PR #40, opened 2021-09-24
Generate password when creating a user

### LINKED ISSUE #6: Generate a random password when creating a new user
In the admin dashboard, when creating a new user, the current behavior allows the admin to create a password, sending it to the right after, we should instead generate a random password in our backend and send it to the user by e-mail.

The endpoint is `/admin/users/new`

image

marmitoTH:
Close #6 

The password field has been removed from the user creation form, it's now randomly generated with [passgen](https://github.com/cryptice/Passgen), respecting Devise's minimum password length.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,8 @@ gem 'aasm', '~> 5.0', '>= 5.0.8'
 gem 'platform-api', github: 'heroku/platform-api'
 gem 'letsencrypt-rails-heroku', group: 'production'
 
+gem 'passgen'
+
 group :test do
   gem 'capybara'
   gem 'shoulda-matchers'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -317,6 +317,7 @@ GEM
     passenger (6.0.8)
       rack
       rake (>= 0.8.1)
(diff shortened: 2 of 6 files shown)


Gemfile
```diff
@@ -61,6 +61,8 @@ gem 'aasm', '~> 5.0', '>= 5.0.8'
 gem 'platform-api', github: 'heroku/platform-api'
 gem 'letsencrypt-rails-heroku', group: 'production'
 
+gem 'passgen'
```
adbatista, line 64: you don't need a gem for this Ruby provides SecureRandom you can use it

adbatista (changes requested):

Gemfile
marmitoTH: SecureRandom is cool, but it just generates stuff like `hex`, `base64`, `uuid`. I guess it would be good enough for a temporary password, but passgen just generates strong passwords out of the box, with no specific patterns, it's better IMHO.

adbatista:
We decided to go in another way so this PR will not be necessary anymore thanks for your contribution :D

Issue #8, opened 2021-08-30
Apply a better design in recovering password page


wenderjean:
We should fix the design of the recovering password page in order to make it similar to the sign-in one.

![image](https://user-images.githubusercontent.com/941776/131349033-98b669f8-3e49-4a76-8392-66ea2587a005.png)

PR #43, opened 2021-10-04
[8] Applies a better design in recovering password page

LucasHFS:
solves #8.


config/locales/pt-BR/pt-BR.yml
```diff
@@ -88,3 +88,13 @@ pt-BR:
...
+  form:
+    title:
+      change_your_password: 'Mude sua senha'
+    hint:
+      minimum_caracters: 'caracteres mínimos'
+    field:
+      new_password: 'Nova Senha'
+      confirm_new_password: 'Confirmar nova senha'
+    button:
+      change_my_password: 'Mudar minha senha'
```
adbatista, line 100: configure your editor to add new line in the end of file or add editorconfig

wenderjean (commented):
Even though it is not the best design experience I'd rather much use a simpler approach like that one applied to the user profile page, the current approach (with this PR) is very similar to the log-in page.

LucasHFS:
I Applied the suggested approach, that is similar to the user profile page

image

PR #62, opened 2021-11-17
Modify the message in the notification mail to user fill punchs

NondasNF:
Simple modify in the mail body and respective spec of "notification user fill punchs" feature of mailer service.
Change the position of the numbers in the body of the email, placing them in ascending order

### DIFF
--- a/config/locales/pt-BR/views.yml
+++ b/config/locales/pt-BR/views.yml
@@ -27,7 +27,7 @@ pt-BR:
         Pessoal,
       </p>
       <p>
-        Preencham o punch entre os dias 16 e 15.
+        Preencham o punch entre os dias 15 e 16.
       </p>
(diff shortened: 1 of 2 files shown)


wenderjean:
Would you mind creating a related issue to this change?

config/locales/pt-BR/views.yml
```diff
@@ -27,7 +27,7 @@ pt-BR:
         Pessoal,
       </p>
       <p>
-        Preencham o punch entre os dias 16 e 15.
+        Preencham o punch entre os dias 15 e 16.
```
wenderjean, line 30: I believe we can improve this message adding months to it, something like: `Preencham o punch entre os dias 15/<last month> e 16/<current month>`.

PR #65, opened 2021-11-18
Feature/separate two factor form from sign in

### LINKED ISSUE #12: Add a second step sign-in process
The current sign-in form always shows the 2FA field, new users are always confused about what to fill up in that field.

I remind when the feature has been created the reason to let the field always be present was the security risk we would create performing an ajax request to ask the backend if the user has/hasn't the 2FA enabled.

A better approach to fix the confusion regarding this field and also the security risk is to work in a two-step sign-in like almost all services do, if the 2FA is enabled for the user when they reach `Sign In` button we would redirect to another form to validate 2FA.

cc/ @adbatista

marmitoTH:
Closes #12 

The two factor OTP code input now have it's own view, and it only shows up when the user have 2FA enabled.

<details>
<summary>Screenshots</summary>

image
image
</details>


app/controllers/concerns/two_factor_authentication.rb
```diff
@@ -0,0 +1,48 @@
...
+    if @user.valid_password?(user_params[:password])
+      session[:otp_user_id] = @user.id
+      render 'devise/sessions/two_factor'
+    end
+  end
+
+  def authenticate(user)
+    session.delete(:otp_user_id)
+    remember_me(user) if user_params[:remember_me]
+    user.save!
```
wenderjean, line 28: Is there a specific issue that requires us to reach `save!` here?

app/controllers/concerns/two_factor_authentication.rb
```diff
@@ -0,0 +1,48 @@
+module TwoFactorAuthentication
```
wenderjean, line 1: What do you think about favoring a better separation of concerns by creating a service to deal with the user stuff and letting this module deal with session/redirection stuff only?

Example:
```ruby
module TwoFactorAuthentication
  extend ActiveSupport::Concern

  def authenticate_otp
    VerifyUserAuthencity.call(**verify_user_authenticity_params) do |result|
      @user = result.user

      return unless result.otp_required_for_login?
      return assign_session_and_redirect(@user) if result.authenticate?

      authenticate(@user) if result.valid_otp?
    end
  end

  private

  def verify_user_authenticity_params
    {
      email: user_params[:email],
      password: user_params[:password],
      user_id: session[:otp_user_id],
      otp_attempt: user_params[:otp_attempt]
    }
  end

  def assign_session_and_redirect(user)
    session[:otp_user_id] = user.id

    render 'devise/sessions/two_factor'
  end

  def authenticate(user)
    session.delete(:otp_user_id)

    remember_me(user) if user_params[:remember_me]

    sign_in(user, event: :authentication)
  end

  def user_params
    params.require(:user).permit(:email, :password, :remember_me, :otp_attempt)
  end
end
```

wenderjean: Your service would look like:

```ruby
class VerifyUserAuthencity
  Result = Struct.new(:user, :otp_attempt, :password) do
  	def otp_required_for_login?
  	  user.otp_required_for_login?
  	end

    def authenticate?
      !otp_attempt && user.valid_password?(password)
    end

    def valid_otp?
      user.validate_and_consume_otp!(otp_attempt) ||
        user.invalidate_otp_backup_code!(otp_attempt)
    end
  end

  NilResult = Struct.new do
  	def otp_required_for_login?
  	  return false
  	end

    def authenticate?
      return false
    end

    def valid_otp?
      return false
    end
  end

  def self.call(email:, password:, user_id:, otp_attempt:, &block)
  	new(email, password, user_id, otp_attempt).call(&block)
  end

  def initialize(email, password, user_id = nil, otp_attempt = nil)
  	@email = email
  	@password = password
  	@user_id = user_id
  	@otp_attempt = otp_attempt
  end

  def call
  	if (user = find_user)
  	  yield Result.new(user, @otp_attempt, @password)
  	else
      yield NilResult.new
  	end
  end

  private

  def find_user
  	if @email
      User.find_by(email: @email)
    elsif @user_id
      User.find(@user_id)
    end
  end
end
```

wenderjean: by doing that we're favoring separation of concerns, avoiding nils returning result objects, and also favoring a simpler/testable solution.

marmitoTH: It does look much better, but I can't generate structs without attributes for the `NilResult`. Do you have any alternative for this?

wenderjean: @marmitoTH I see two different options but I think it's also good as is.
1. Create real classes (too much complexity probably)
2. Use a single Result object and add safe navigator to all methods like `user&.otp_required_for_login?`

wenderjean (approved):
Thanks for addressing the changes. LGTM

adbatista (approved):

adbatista:
@marmitoTH please rebase your code

marmitoTH:
@adbatista Why? For me it's saying that there's no conflicts with the base branch.

adbatista:
@marmitoTH CI is Broken

Issue #71, opened 2021-11-29
Create UX design system using Tailwind


wenderjean:
We decided to modernize our application using new tools and approaches for UI/UX, for that reason we want to create a new design system using [tailwindcss](https://tailwindcss.com/). The main idea here is to perform the setup for Tailwind and also create a minimal set of components to be used for the subsequent redesign tickets.

Suggestion:

- https://bradfrost.com/blog/post/atomic-web-design/

Related ticket:

- https://github.com/Codeminer42/Punchclock/issues/61

tauantcamargo:
let me get this one 😄

lukzgois:
That's interesting 😄 . I'm creating the new screens using a component-driven design approach, so, the components are already separated in the Figma file (I've attached it in the ticket #61). I think you can take a look on it @tauantcamargo, maybe starting to create the components for the new screens.

@wenderjean @adbatista I can create some tickets for each of these new components. What do you think about it?

adbatista:
> @adbatista I can create some tickets for each of these new components. What do you think about it?

that will help, thanks

PR #74, opened 2021-12-06
chore: adding and configuring tailwind to project

tauantcamargo:
Closes [#71](https://github.com/Codeminer42/Punchclock/issues/71)

Was added tailwind to project and made a default configuration.


Gemfile
```diff
@@ -57,6 +57,8 @@ gem 'active_skin'
 gem 'active_admin_flat_skin'
 gem 'aasm', '~> 5.0', '>= 5.0.8'
 
+gem 'turbolinks', '~> 5.2.0'
```
adbatista, line 60: why?

tauantcamargo: i removed, i was following an implementation but i forgot to remove it, thank u.

tailwind.config.js
```diff
@@ -0,0 +1,172 @@
+module.exports = {
```
adbatista, line 1: did you get these configs from the Lucas's design?

fabiosammy: @tauantcamargo :point_up: ?

Issue #70, opened 2021-11-29
The "User" field contains all users (contributions creation page)


wenderjean:
When creating a new contribution (manually) the field "Users" contains all users but It should contains only active engineers, the result is not ordered also.

https://github.com/Codeminer42/Punchclock/blob/master/app/admin/contributions.rb#L9

PR #76, opened 2021-12-07
Corrects user select collection

LucasHFS:
Solves #70

### DIFF
--- a/spec/queries/company_users_collection_query_spec.rb
+++ b/spec/queries/company_users_collection_query_spec.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe CompanyUsersCollectionQuery do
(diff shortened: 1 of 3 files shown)


spec/queries/company_users_collection_query_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    let(:company2) { create(:company) }
+    let(:user3) { create(:user, company: company2, name: 'userA') }
+    let(:user4) { create(:user, company: company2, name: 'userB') }
+
+    before do
+      user1
+      user2
+      user3
+      user4
```
adbatista, line 21: if you need the user to be created before running the test use let! instead https://rspec.info/documentation/3.10/rspec-core/RSpec/Core/MemoizedHelpers/ClassMethods.html#let!-instance_method

PR #80, opened 2022-01-03
FIx CVE-2021-44528 - upgrade actionpack

wenderjean:
Related to: #79


spec/services/qr_code_generator_service_spec.rb
```diff
@@ -25,11 +25,11 @@
 
   it 'create QR Code with the right URI' do
     image_data = QrCodeGeneratorService.call(user)
-    
+
     expect(RQRCode::QRCode)
       .to have_received(:new)
-      .with('otpauth://totp/Punchclock:example@exp.com?secret=fake_secret&issuer=Punchclock')
-      
+      .with('otpauth://totp/Punchclock:Punchclock_example%40exp.com?secret=fake_secret&issuer=Punchclock')
```
wenderjean, line 31: Guys from `rotp` gem decided to implement a requirement from Google 2FA repeating the issuer in the URL, not sure yet but I believe this change will require everyone to remove/add the URL again.

https://github.com/mdp/rotp/pull/51

PR #89, opened 2022-02-21
Create punch attachment tests

andreLumor:
The attachment link rendering and its click action weren't being tested.
To test this:
1. Created a with_attachment trait in the punch factory.
2. Created attachment fixture.
3. Instantiated a punch with attachment before tests.
4. Tested if punch.attachment_url was rended on the page.
5. Tested if attachment content was shown after click.

### DIFF
--- a/spec/features/punches_list_spec.rb
+++ b/spec/features/punches_list_spec.rb
@@ -10,6 +10,9 @@ feature 'Punches list' do
   let!(:other_punch) {
     create(:punch, user_id: authed_user.id, company_id: authed_user.company_id)
   }
+  let!(:punch_with_attachment) {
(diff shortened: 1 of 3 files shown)


spec/features/punches_list_spec.rb
```diff
@@ -48,4 +51,13 @@
...
     expect(page).to have_css('tbody tr', count: 1)
   end
+
+  it "Renders attachment link" do
+    expect(page).to have_link('', href: punch_with_attachment.attachment_url)  
+  end
+
+  scenario 'follow attachment link' do
+    find("a[href='#{punch_with_attachment.attachment_url}']").click
+    expect(page).to have_content('This is a punch attachment')
```
andreLumor, line 61: Couldn't find a good way to get attachment body, so i just transcribed it.
Is there a better way to do this?

PR #96, opened 2022-03-22
Improves tests for ContributionsByUserQuery

mateuscelio:
* The old test fails when running in March, returning 6 contributions instead of the expected 3
* Changes descriptions to third-person singular
* Moves subject to the first position after `context` blocks

### DIFF
--- a/spec/queries/contributions_by_user_query_spec.rb
+++ b/spec/queries/contributions_by_user_query_spec.rb
@@ -8,22 +8,24 @@ RSpec.describe ContributionsByUserQuery do
   let(:user)    { create(:user, company: company, office: offices.first) }
 
   context '#to_hash' do
+    subject(:to_hash) { described_class.new.to_hash }
+
     before do
       create_list(:contribution, 6, { user: user, company: company })
     end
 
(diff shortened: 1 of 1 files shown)


spec/queries/contributions_by_user_query_spec.rb
```diff
@@ -45,12 +45,15 @@
   end
 
   context '#per_month' do
+    let(:today) { Date.today }
+    let(:last_month) { today.month - 1 }
```
fabiosammy, line 49: Are you sure about it? If the month is January, what will be the `last_month` value?

What you can do is `(today - 1.month).month`, but again, for January, will it generate the expected behavior?

mateuscelio: https://github.com/Codeminer42/Punchclock/pull/96/commits/91fa81c0d2c9e1b8754d6e4846e01fda60dd299c

wenderjean (approved):

adbatista:
there's conflicts on your code

wenderjean (approved):

PR #97, opened 2022-03-25
Separate evaluations on view by performance and english

GabrielRMuller:
On the evaluations page we would only show the last evaluation done by the user; This is not ideal because we have more than one kind of evaluation and I'd would like to see the last one made for each of them.

This PR separates on the evaluations index view by `performance` and `english`.

Current version:
image

Suggestion:
image


app/views/evaluations/index.html.erb
```diff
@@ -25,12 +27,19 @@
                   <td><%= user.level %></td>
                   <td><%= user.office_city%></td>
                   <td>
-                    <% unless user.evaluations.last.nil? %>
-                      <%= user.last_evaluation.evaluation_date || user.last_evaluation.created_at.strftime('%e/%m/%Y') %> -
-                      <%= link_to "View", evaluation_path(user.last_evaluation.id)%>
+                    <% unless user.last_performance_evaluation.nil? %>
```
matheusazzi, line 30: That seems easier to reason about.
```suggestion
                    <% if user.last_performance_evaluation %>
```

app/views/evaluations/index.html.erb
```diff
@@ -25,12 +27,19 @@
...
+                      <%= user.last_performance_evaluation.evaluation_date || user.last_performance_evaluation.created_at.strftime('%e/%m/%Y') %> -
+                      <%= link_to "View", evaluation_path(user.last_performance_evaluation.id)%>
                     <% end%>
                   </td>
-                  <td><%= user.last_evaluation.score %></td>
+                  <td><%= user.last_performance_evaluation&.score %></td>
+                  <td>
+                    <% unless user.last_english_evaluation.nil? %>
+                      <%= user.last_english_evaluation.evaluation_date || user.last_english_evaluation.created_at.strftime('%e/%m/%Y') %> -
```
matheusazzi, line 38: A tip for a follow-up PR would be creating an Evaluation [viewmodel/presenter/decorator](https://www.nopio.com/blog/view-objects-rails-views/) to abstract complex logic like this:
```ruby
user.last_english_evaluation.evaluation_date || user.last_english_evaluation.created_at.strftime('%e/%m/%Y')
```

Also, we can leverage [Null Object Pattern](https://medium.com/@kelseydh/using-the-null-object-pattern-with-ruby-on-rails-b645ebf79785) there, so we can get rid of those `.try`s and `||`s, having less defensive programming.

app/views/evaluations/index.html.erb
GabrielRMuller: it really does, good catch! :smile:

spec/support/capybara.rb
```diff
@@ -4,7 +4,7 @@
 
 Capybara.register_driver :headless_chrome do |app|
   capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
-    chromeOptions: {  args: %w[no-sandbox headless disable-gpu window-size=1280,1024] }
+    "goog:chromeOptions": {  args: %w[no-sandbox headless disable-gpu window-size=1280,1024] }
```
GabrielRMuller, line 7: `Resolved issue 1786: Feature request: change chromeOptions to goog:chromeOptions [['Pri-3']]`
see: https://chromedriver.storage.googleapis.com/2.31/notes.txt

adbatista (approved):

Issue #106, opened 2022-04-20
Create an endpoint to register a punch on API


alexandresaura:
We need to create an endpoint that will register a punch to the authenticated user.

alexandresaura:
Could I take this one?

PR #110, opened 2022-04-27
Add API create punch endpoint

alexandresaura:
Issue: https://github.com/Codeminer42/Punchclock/issues/106

In this PR we are adding the punches create endpoint that will receive a body in this format:
```json
{
  "punches": {
    "2022-04-19": [
      {
        "from": "2022-04-19T12:00:00.000Z",
        "to": "2022-04-19T15:00:00.000Z",
        "project_id": "3"
      },
      {
        "from": "2022-04-19T17:00:00.000Z",
        "to": "2022-04-19T21:00:00.000Z",
        "project_id": "3"
      }
    ]
  }
}
```
Will delete all punches in the same day and create this new ones


app/controllers/api/v1/punches_controller.rb
```diff
@@ -0,0 +1,29 @@
+module Api
+  module V1
+    class PunchesController < ApiController
+      def create
+        new_punches = params[:punches]
+        return render json: { message: 'There are no punches to create' }, status: :bad_request if new_punches.nil? || new_punches.values.empty?
```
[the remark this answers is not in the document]
alexandresaura: Thanks, I didnt know this http status code

spec/controllers/api/v1/punches_controller_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+                "from": "2022-04-19T12:00:00.000Z",
+                "to": "2022-04-19T15:00:00.000Z",
+                "project_id": "#{project.id}"
+              },
+              {
+                "from": "2022-04-19T16:00:00.000Z",
+                "to": "2022-04-19T21:00:00.000Z",
+                "project_id": "#{project.id}"
+              }
+            ]
```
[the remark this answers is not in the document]
alexandresaura: I did it this way because I used it based on another code about punches in this repository, but it makes more sense. Thanks

spec/controllers/api/v1/punches_controller_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+          }
+        }
+      }
+
+      subject { post :create, params: params }
+
+      before do
+        request.headers.merge(headers)
+      end
```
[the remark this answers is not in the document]
alexandresaura: I think in this version does not have this property:
image

app/controllers/api/v1/punches_controller.rb
```diff
@@ -0,0 +1,29 @@
+module Api
+  module V1
+    class PunchesController < ApiController
+      def create
```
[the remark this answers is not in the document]
alexandresaura: Thanks, I did it, but I created a service object with an static function that can create a punch, I dont know if will be better create a Instance of a service and use that after.

spec/controllers/api/v1/punches_controller_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+      it 'returns content type json' do
+        expect(subject.content_type).to eq 'application/json; charset=utf-8'
+      end
+
+      it 'returns a missing punches message' do
+        expect(JSON.parse(subject.body)).to eq('message' => 'There are no punches to create')
+      end
+    end
+
+    context 'when api token is valid, but does not have params' do
```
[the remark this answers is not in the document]
alexandresaura: Cool, I didnt know about I will implement

spec/controllers/api/v1/punches_controller_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+  let(:company) { create(:company) }
+  let(:user) { create(:user, :with_token, company: company) }
+  let(:project) { create(:project, company: user.company) }
+  let(:headers) { { token: user.token } }
+
+  describe 'POST api/v1/punches' do
+    context 'when api token is valid and params are valid' do
+      let(:params) {
+        {
+          "_json": [
```
rhian-cs, line 13: Instead of calling this main key `_json`, you could call it `punches` for better readability.

spec/controllers/api/v1/punches_controller_spec.rb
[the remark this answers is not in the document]
rhian-cs: I see, in that case I don't know of any other options.

app/controllers/api/v1/punches_controller.rb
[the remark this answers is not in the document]
rhian-cs: Both patterns work, but if you want to use instance variables you may want to use an instance of a service instead.

spec/controllers/api/v1/punches_controller_spec.rb
rhian-cs, line 13: You could call this key `punches` for better readability.

app/services/punches_service.rb
```diff
@@ -0,0 +1,15 @@
+class PunchesService
+  def self.create(punches, user)
+    deletes = punches.map { |punch| punch['from'] }
+
+    @punches = user.punches
```
rhian-cs, line 5: I don't think you need to use instance variables here, since you're not using them in the view layer.

app/services/punches_service.rb
```diff
@@ -0,0 +1,15 @@
+class PunchesService
+  def self.create(punches, user)
+    deletes = punches.map { |punch| punch['from'] }
```
rhian-cs, line 3: You could call this variable something like `dates_to_delete` or `punch_dates_to_delete` for better readability.

app/services/punches_service.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    @punches = user.punches
+    created_punches = @punches.transaction do
+      @punches.by_days(deletes).delete_all
+      @punches.where(
+        company: user.company
+      ).create(punches)
+    end
```
rhian-cs, line 11: You could put this block in a separate method for better readability (e.g. `create_punches(punches)`).

app/services/punches_service.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    return created_punches
```
rhian-cs, line 13:
```suggestion
```
Since this value is already returned by the previous block, you don't need this explicit return.

spec/controllers/api/v1/punches_controller_spec.rb
alexandresaura: When I send a request with the params:
```json
[
  {
    "from": "2022-04-19T12:00:00.000Z",
    "to": "2022-04-19T15:00:00.000Z",
    "project_id": "3"
  },
  {
    "from": "2022-04-19T17:00:00.000Z",
    "to": "2022-04-19T21:00:00.000Z",
    "project_id": "3"
  }
]
```
The rails put its values on a params called _json, but when I send by test, I need to declare this key

app/services/punches_service.rb
alexandresaura: Cool, I didnt know this instance variables in ruby, I thought it was a special variable for transactions

app/services/punches_service.rb
alexandresaura: I will do it

app/services/punches_service.rb
alexandresaura: Thanks, I will create another method for this

app/services/punches_service.rb
```diff
@@ -0,0 +1,20 @@
+class PunchesService
```
wenderjean, line 1: What do you think about naming this service `CreatePunchService`? It sounds more like a service object.

app/services/punches_service.rb
```diff
@@ -0,0 +1,20 @@
...
+  def self.create(punches, user)
```
wenderjean, line 2: This method is kinda anemic at the same time you have another method doing two different things, what do you think about keeping specialized methods? Suggestion:

```ruby
class CreatePunchService
  class << self
    def call(collection_of_punches, user)
      days_to_delete = collection_of_punches.map { |punch| punch['from'] }

      ActiveRecord::Base.transaction do
        delete_punches_for_the_same_day(user, days_to_delete)
        create_new_punch(user, collection_of_punches)
      end
    end

    private

    def delete_punches_for_the_same_day(user, days_to_delete)
      user.punches.by_days(days_to_delete).delete_all
    end

    def create_new_punch(user, collection_of_punches)
      users.punches.where(
        company: user.company
      ).create(collection_of_punches)
    end
  end
end
```

app/services/punches_service.rb
```diff
@@ -0,0 +1,20 @@
...
+    create_delete_punches(punches, punches_date_to_delete, user)
+  end
+
+  private
+
+  def self.create_delete_punches(punches, punches_date_to_delete, user)
+    users_punches = user.punches
+
+    users_punches.transaction do
+      users_punches.by_days(punches_date_to_delete).delete_all
```
wenderjean, line 14: Is intentional the use of `.delete_all`? `destroy_all` and `delete_all` behave different.

https://spartchou.gitbooks.io/ruby-on-rails-basic/content/syntax/delete_all_vs_destroy_all.html

spec/services/punches_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          },
+          {
+            "from" => "2022-04-19T16:00:00",
+            "to" => "2022-04-19T21:00:00",
+            "project_id" => "#{project.id}"
+          }
+        ]
+      }
+
+      let(:office) { create(:office, :with_holiday) }
```
wenderjean, line 25: You're not using this `office`

spec/services/punches_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      it "returns the created punches" do
```
wenderjean, line 27: What do you think about changing this to something like?
```ruby
it "returns the created punches" do
  result = described_class.create(punches, user)

  # You may need to something like `result.map(&:to_json)`
  expect(result).to match_array(punches)
end
```

spec/services/punches_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+        created_punches = described_class.create(punches, user).map { |punch| {from: punch['from'].strftime("%FT%T"), to: punch['to'].strftime("%FT%T"), project_id: punch['project_id'].to_s } }
+        expected_punches = punches.map { |punch| {from: punch["from"], to: punch["to"], project_id: punch["project_id"].to_s } }
+        expect(created_punches).to match_array(expected_punches)
+      end
+    end
+
+    context 'when does have another punch on same date' do
```
wenderjean, line 34: I believe by using a bit more FactoryBot you'd make your spec more simple.
```ruby
context 'when does have another punch on same date' do
  let!(:punch_1) { create(:punch, from: '2022-04-19T12:00:00', to: '2022-04-19T15:00:00', project: project) }
  let!(:punch_2) { create(:punch, from: '2022-04-19T16:00:00', to: '2022-04-19T21:00:00', project: project) }

  let(:punches) do
    [
      {
        "from" => "2022-04-19T10:00:00",
        "to" => "2022-04-19T12:00:00",
        "project_id" => project.id
      },
      {
        "from" => "2022-04-19T15:00:00",
        "to" => "2022-04-19T20:00:00",
        "project_id" => project.id
      }
    ]
  end

  it 'returns created punches and deletes the last punches' do
    result = described_class.create(punches, user)

    expect(Punch.count).to eq(2)
    expect(result).to match_array(punches)
  end
end
 ```

File: spec/controllers/api/v1/punches_controller_spec.rb [hunk 17]
```diff
@@ -0,0 +1,155 @@
...
+            {
+              "from": "2022-04-19T12:00:00.000Z",
+              "to": "2022-04-19T15:00:00.000Z",
+              "project_id": "#{project.id}"
+            },
+            {
+              "from": "2022-04-19T16:00:00.000Z",
+              "to": "2022-04-19T21:00:00.000Z",
+              "project_id": "#{project.id}"
```
wenderjean, line 22: You don't need to use concatenation format here, you can just write `project.id`

spec/controllers/api/v1/punches_controller_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      end
+
+      it { is_expected.to have_http_status(:created) }
+
+      it 'returns content type json' do
+        expect(subject.content_type).to eq 'application/json; charset=utf-8'
+      end
+
+      it 'returns created punches' do
+        created_punches = JSON.parse(subject.body).map { |punch| {from: punch['from'], to: punch['to'], project_id: punch['project_id'].to_s } }
```
wenderjean, line 41: You shouldn't need to remap everything :thinking: 
Does this work?
```
expect(response.body).to match_array(params[:_json])
```

wenderjean (changes requested):

spec/controllers/api/v1/punches_controller_spec.rb
rhian-cs: I see, perhaps that's the default behavior when you send a JSON array instead of an object.

But what I meant is to make the payload expect a `punches` key, so that the payload would be something like:
```json
{
  "punches": {
    "from": "2022-04-19T12:00:00.000Z",
    "to": "2022-04-19T15:00:00.000Z",
    "project_id": "3"
  },
  {
    "from": "2022-04-19T17:00:00.000Z",
    "to": "2022-04-19T21:00:00.000Z",
    "project_id": "3"
  }
}
```

You could then load the data from inside the controller by using:
```rb
def punch_params
  params.require(:punches).permit(punches: [:from, :to, :project_id])
end
```

app/services/punches_service.rb
alexandresaura: I used the same method that are being used on app/controllers/dashboard_controller.rb

app/controllers/api/v1/punches_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+        new_punches = params[:_json]
+        return render json: { message: 'There are no punches to create' }, status: :unprocessable_entity if new_punches.nil? || new_punches.empty?
```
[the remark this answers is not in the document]
alexandresaura: If I dont do that, when I use the require methods on params It erases an error if does not have punches

app/controllers/api/v1/punches_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+
+        permitted_punches = params.permit(_json: [:from, :to, :project_id]).require(:_json)
```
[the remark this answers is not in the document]
alexandresaura: Thanks, I will do that

app/services/punches_service.rb
alexandresaura: I didnt know in the ruby if a service do one thing or several related with the model

spec/controllers/api/v1/punches_controller_spec.rb
alexandresaura: I will remove, thx

spec/controllers/api/v1/punches_controller_spec.rb
alexandresaura: The response.body has more properties than the params that I send

app/services/punches_service.rb
alexandresaura: Cool! This blow up my mind kkkkk, I didnt know how to stay with the transactions block with 2 methods

app/services/punches_service.rb
wenderjean: Ideally your service is going to perform a single action, that's the reason in general you'll see the service name using the verb as a prefix, `Create<something>`, `Update<something>`

spec/services/punches_service_spec.rb
alexandresaura: I will remove, thx

spec/services/punches_service_spec.rb
alexandresaura: I will refactor this

app/services/punches_service.rb
wenderjean: These two methods are going to be part of the transaction considering you're invoking them inside the transaction block, anyone that raises an error will trigger the rollback action.

app/services/punches_service.rb
alexandresaura: I get it, the problem was in this repo don't have much services with this action prefix

spec/services/punches_service_spec.rb
alexandresaura: Yes, It makes more simple, I will refactor this

wenderjean (dismissed):
LGTM

spec/controllers/api/v1/punches_controller_spec.rb
alexandresaura: Yes, I did it, but the require must be after permit, because the require returns an array and does not have the permit method on array

app/controllers/api/v1/punches_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+    class PunchesController < ApiController
+      def create
+        created_punches = CreatePunchesService.call(create_punches_params, current_user)
+        render json: created_punches, status: :created
+      end
+
+      private
+
+      def create_punches_params
+        params.permit(punches: [:from, :to, :project_id]).require(:punches)
```
adbatista, line 12: you can do params.require(:punches).permit([:from, :to, :project_id])

app/services/create_punches_service.rb
```diff
@@ -0,0 +1,24 @@
+class CreatePunchesService
+  class << self
+    def call(collection_of_punches, user)
+      days_to_delete = collection_of_punches.map { |punch| punch['from'] }
+
+      ActiveRecord::Base.transaction do
+        delete_punches_for_the_same_day(user, days_to_delete)
```
adbatista, line 7: why?

alexandresaura: I use this approach because is the one that is being used on the web.
On the mobile, you will select the date and create both punches, before and after lunch. So we need to guarantee that does not have any other punch on the same day.

adbatista: ok, the action is to create punch and not to delete them, if there's a collision  it should return a validation error

app/controllers/api/v1/punches_controller.rb
adbatista: sorry I showed the wrong syntax it should be `params.require(:punches).permit(:from, :to, :project_id)`

adbatista: https://github.com/rails/strong_parameters#strong-parameters

app/services/create_punches_service.rb
alexandresaura: But, that isn't an endpoint to delete punches, and for now, if the app creates a punch that is wrong it should stay until we create the delete endpoint?

app/controllers/api/v1/punches_controller.rb
wenderjean: The approach is correct, in fact, he's using `require` only to return the value into the "punches" attributes. :stuck_out_tongue:

app/services/create_punches_service.rb
alexandresaura: This delete was removed, instead if there are any punch on the same date, an error will be returned

app/services/bulk_punches_service.rb
```diff
@@ -0,0 +1,22 @@
...
+    private
+
+    def there_punch_on_the_same_day?(user, insert_days)
+      user.punches.by_days(insert_days).present?
+    end
+
+    def create_new_punch(user, collection_of_punches)
+      user.punches.where(
+        company: user.company
+      ).create(collection_of_punches)
```
fabiosammy, line 19: I'm thinking about what will happen if you send a collection without the `project_id` or `to` data. , and why are you filtering the punches with the user company, since you are getting the punches of the current user?

app/controllers/api/v1/punches_controller.rb
```diff
@@ -5,11 +5,22 @@ def index
         render json: scoped_punches, status: :ok, each_serializer: PunchSerializer
       end
 
+      def bulk
+        created_punches = BulkPunchesService.call(bulk_punches_params, current_user)
+        render json: created_punches, status: :created, each_serializer: PunchSerializer
+      rescue StandardError => e
```
fabiosammy, line 11: I do not like this type of approach. Avoid using `rescue` to control your data flow.
The `rescue` can generate some unexpected behavior.

app/services/bulk_punches_service.rb
```diff
@@ -0,0 +1,22 @@
+class BulkPunchesService
```
fabiosammy, line 1: The concept to use a service object is just to take a single action, and process the data.
But for your case, you need to validate your data too.
Prefer to use the Service Object pattern when you are ensured about your data, and avoid it when you're passing params data from a user request.

Take a look at the Form Object pattern. An inspiration: https://jaryl.medium.com/disciplined-rails-form-object-techniques-patterns-part-2-12b8d530143d

app/services/bulk_punches_service.rb
alexandresaura: Good catch, I removed that and create the punches instances before saving now, so I don't need that.

app/controllers/api/v1/punches_controller.rb
alexandresaura: I didn't think in this way, any error can be sent to the rescue and it should be a problem, thanks.

app/services/bulk_punches_service.rb
alexandresaura: Thanks, I didn't know about this pattern, so I implemented it.

app/forms/punches_create_form.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  def punch_on_same_day_validation
+    @punches.each do |punch|
+      errors.add(:from, 'There is already a punch on the same day') if there_punch_on_the_same_day?(punch)
+    end
+  end
+
+  def there_punch_on_the_same_day?(punch)
```
rhian-cs, line 43:
```suggestion
  def multiple_punches_on_the_same_day?(punch)
```
I think this method could have a different name for better readability.

app/forms/punches_create_form.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  attr_accessor :punches
+
+  validates :punches, presence: true
+  validate :punches_validations, :punch_on_same_day_validation
+
+  def initialize(params)
+    params ||= {}
+    params[:punches] ||= []
+    params[:user] ||= User.new
```
rhian-cs, line 12:
```suggestion
  def initialize(punches: [], user: User.new)
```
You can use the keyword syntax to reduce some code.

This will still expect a hash to be passed, but your method already has access to the variables directly.

app/forms/punches_create_form.rb
```diff
@@ -0,0 +1,46 @@
...
+      Punch.new(**punch_params, user: params[:user], company: params[:user].company)
+    end
+  end
+
+  def save
+    return false unless valid?
+
+    @punches = Punch.transaction do
+      @punches.each(&:save!)
+    end
```
rhian-cs, line 24: [Question] Should this really be assigning to `@punches`?

rhian-cs (commented):
Just some minor comments, I'd say the code is looking pretty good.

adbatista:
what is missing here?

app/forms/punches_create_form.rb
alexandresaura: Thanks, I didnt know about that

app/forms/punches_create_form.rb
alexandresaura: I don't thing this is the idea, the function wanna know if there are at least 1 other punch already on database, not if have multiples

app/forms/punches_create_form.rb
alexandresaura: I thought that I needed that to reload, for example the id, but I don't need, thanks for the observation.

app/forms/punches_create_form.rb
rhian-cs: In that case you can call it `has_punches_on_the_same_day?`

Also, the function itself doesn't check for a single punch, but for multiple punches (which is the correct validation), so I think using `punches` instead of `punch` in the method name makes more sense.

alexandresaura:
@adbatista I only need some reviews, but I think it's all right.

app/forms/punches_create_form.rb
```diff
@@ -0,0 +1,42 @@
+class PunchesCreateForm
+  include ActiveModel::Model
+
+  attr_accessor :punches
+
+  validates :punches, presence: true
+  validate :punches_validations, :punch_on_same_day_validation
+
+  def initialize(punches: [], user: User.new)
```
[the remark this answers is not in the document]
adbatista: ok, but the user should not come from the request, and the initialize will aways receive an arg so should not care  about this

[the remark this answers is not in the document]
alexandresaura: Ok, I removed

adbatista (approved):

Issue #114, opened 2022-04-29
Order repository component by the number of contributions when manually creating a new contribution


wenderjean:
When adding a new contribution manually using `admin > contribution > new contribution` is almost impossible to find the correct repository in the select component.

![image](https://user-images.githubusercontent.com/941776/166007536-c867d077-2367-426e-822d-dc9a81565c09.png)

PR #115, opened 2022-05-05
Update repository select to show repositories ordered by total of contributions

EduardoSCosta:
Resolve: [#114 ](https://github.com/Codeminer42/Punchclock/issues/114)

- [x] When creating a new contribution, the repositories should be ordered by their number of contributions.

Screenshot_20220505_105545 (1)

### DIFF
--- a/app/queries/repositories_ordered_by_contributions_query.rb
+++ b/app/queries/repositories_ordered_by_contributions_query.rb
@@ -1,0 +1,15 @@
+class RepositoriesOrderedByContributionsQuery
+  def call
+    repositories_query = <<-SQL
+      SELECT r.id, r.link FROM repositories r
+        ORDER BY (
+          SELECT count(*)
+          FROM contributions c
+          WHERE r.id = c.repository_id
+        ) DESC
+    SQL
+
+    repositories_rank = ActiveRecord::Base.connection.execute(repositories_query)
+    repositories_rank.map { |repository| [repository['link'], repository['id']] }
+  end
+end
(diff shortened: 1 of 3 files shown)


app/queries/repositories_ordered_by_contributions_query.rb
```diff
@@ -0,0 +1,13 @@
+class RepositoriesOrderedByContributionsQuery
+  def call
+    repositories_query = "SELECT r.id, r.link FROM repositories r
```
wenderjean, line 3: What do you think about writing this using a HEREDOC format like
```
<<~SQL
  ...
SQL
```

spec/queries/repositories_ordered_by_contributions_query_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    let!(:repo1) { create(:repository) }
+    let!(:repo2) { create(:repository) }
+    let!(:repo3) { create(:repository) }
+
+    let!(:contribution1) { create(:contribution, repository: repo2) }
+    let!(:contribution2) { create(:contribution, repository: repo2) }
+    let!(:contribution3) { create(:contribution, repository: repo3) }
+
+    context 'when called' do
```
wenderjean, line 17: We could improve this spec doing something like
```ruby
context 'when all repositories have the same number of contributions' do

end

context 'when repositories have different number of contributions each' do

end
```

wenderjean (approved):
LGTM

PR #116, opened 2022-05-05
Adds oauth2 provider 

mateuscelio:
Fixes #102

## Adds [OAuth2](https://datatracker.ietf.org/doc/html/rfc6749) provider to Punchclock using [Doorkeeper](https://github.com/doorkeeper-gem/doorkeeper).

* Authorization grants configurated are [authorization code](https://datatracker.ietf.org/doc/html/rfc6749#section-1.3.1) and [client credentials](https://datatracker.ietf.org/doc/html/rfc6749#section-1.3.4)
* Access token is in JWT format with the following payload
```rb
{
      iss: 'Punchclock',
      iat: Time.current.utc.to_i,
      exp: 2.hours.from_now.to_i,
      jti: SecureRandom.uuid,

      user: {
        id: user.id,
        email: user.email
      }
    }
```
* Access tokens expire in 2h  and can be created new ones using refresh token
* Tokens hash are saved in DB instead of in plaintext format
* Protects API routes with access token using header property `Authorization` with `Bearer` prefix
* Old token authentication method was kept 
* Adds OAuth2 view to manage applications that only Super Admin has access
* Adds view for users manage theirs authorized applications access 
* Changes after sign in when an unaunthenticated user gets  `/oauth/authorize` route.

## Authorization grants flows ex

* [Authorization code](https://github.com/doorkeeper-gem/doorkeeper/wiki/Authorization-Code-Flow)
* [Client credentials](https://github.com/doorkeeper-gem/doorkeeper/wiki/Client-Credentials-flow)

## Views

Links to OAuth2 applications manager route
image

OAuth2 Applications manager view
image

Edit/create OAuth2 Application view 
image

Links to user authorized applications manager
image

User  Authorized applications view
image

PR #119, opened 2022-05-23
Draft: Api documentation with OpenAPI/Swagger

DeboraBSilva:
Generating Swagger API documentation with Rspec tests using Rswag gem.


wenderjean:
Blocked by https://github.com/Codeminer42/Punchclock/pull/116

app/controllers/api/v1/companies_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Api
   module V1
-    class CompaniesController < ApiController
+    class CompaniesController < ActionController::API
```
DeboraBSilva, line 3: I changed the class used here as it was not possible to pass a custom authorization on the endpoint with the `ApiController` class. In order to maintain the custom authentication that `ApiController` implements, it is necessary to review the code to work with Swagger.

app/controllers/api/v1/punches_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Api
   module V1
-    class PunchesController < ApiController
+    class PunchesController < ActionController::API
```
DeboraBSilva, line 3: I changed the class used here as it was not possible to pass a custom authorization on the endpoint with the `ApiController` class. In order to maintain the custom authentication that `ApiController` implements, it is necessary to review the code to work with Swagger.

spec/requests/api/v1/companies_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'swagger_helper'
+
+RSpec.describe 'api/v1/companies', type: :request do
+
+  before { create_logged_in_user }
+
+  path '/api/v1/users' do
+    
+    get('list users') do
```
DeboraBSilva, line 9: The way it is now, the endpoint request only works if you login in punch before. Another possible approach is to pass the token as a parameter(maybe in header or in query), it will look something like this: 
`get('list users') do
      parameter name: :token, in: :header, type: :string`

Issue #118, opened 2022-05-23
Evaluations filter does not work for "type of questionnaire" field 


wenderjean:
When you navigate to the "Evaluations" page and try to filter by "type of questionnaire" the filter does not apply.

![image](https://user-images.githubusercontent.com/941776/169833733-56e29743-d4f5-49f5-924e-1ad6ec0c5690.png)

PR #120, opened 2022-05-24
Fix evaluations type of questionnarie filter

alexandresaura:
Fixes: https://github.com/Codeminer42/Punchclock/issues/118

# Evaluations
Fixing some issues related to the filters and adding the tests

## Questionnaire Filter
Now, the select shows the translated name instead of the code value and the filter works.
image

## Tests
And, adding tests for all filters to ensure that all will work.

## Labels
Fix the English label level to show the translated value too.
image

### DIFF
--- a/spec/features/admin/evaluations_spec.rb
+++ b/spec/features/admin/evaluations_spec.rb
@@ -10,38 +10,107 @@ describe 'Admin Evaluation', type: :feature do
   end
 
   describe 'Filters' do
+    let(:evaluator_user) { create(:user, :active_user, company: admin_user.company) }
+    let(:evaluated_user) { create(:user, :active_user, company: admin_user.company) }
+
+    let!(:evaluation1) do
+      create(:evaluation,
+             :english,
+             evaluator: evaluator_user,
+             evaluation_date: 3.weeks.ago,
+             created_at: 5.months.ago,
+             company: admin_user.company)
+    end
+
+    let!(:evaluation2) do
+      create(:evaluation,
+             :performance,
+             evaluated: evaluated_user,
+             evaluation_date: 5.weeks.ago,
+             created_at: 8.months.ago,
+             company: admin_user.company)
+    end
+
     before do
-      create(:evaluation)
       visit '/admin/evaluations'
     end
 
     it 'by evaluator' do
       within '#filters_sidebar_section' do
         expect(page).to have_select('Avaliador', options: admin_user.company.users.pluck(:name) << 'Qualquer')
+
+        select evaluator_user.name, from: 'q_evaluator_id'
+        click_button 'Filtrar'
+      end
+
+      within_table 'index_table_evaluations' do
+        expect(page).to have_css('tbody tr', count: 1)
+        expect(page).to have_selector("#evaluation_#{evaluation1.id}", count: 1)
+        expect(page).not_to have_selector("#evaluation_#{evaluation2.id}")
       end
     end
 
     it 'by evaluated' do
       within '#filters_sidebar_section' do
         expect(page).to have_select('Avaliado', options: admin_user.company.users.pluck(:name) << 'Qualquer')
+
+        select evaluated_user.name, from: 'q_evaluated_id'
+        click_button 'Filtrar'
+      end
+
+      within_table 'index_table_evaluations' do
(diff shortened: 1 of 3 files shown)


spec/features/admin/evaluations_spec.rb
```diff
@@ -10,38 +10,110 @@
   end
 
   describe 'Filters' do
+    let!(:user1) { create(:user,
+                          active: true,
+                          company: admin_user.company) }
+
+    let!(:user2) { create(:user,
+                          active: true,
+                          company: admin_user.company) }
```
rhian-cs, line 19: You could try to fit these into a single line, if it doesn't exceed the line limit.

You could also rename the variables to `evaluated_user` and `evaluator_user` for better readability.

You can also remove the `!` from the `let` declaration, since the users are already called when creating the evaluations, which are declared with a `!`.

spec/features/admin/evaluations_spec.rb
```diff
@@ -10,38 +10,110 @@
...
+
+    let!(:evaluation1) { create(:evaluation,
+                                :english,
+                                evaluated: user1,
+                                evaluator: user2,
+                                evaluation_date: 3.week.ago,
+                                created_at: 5.month.ago,
+                                company: admin_user.company) }
```
rhian-cs, line 27:
```suggestion
    let!(:evaluation1) do
      create(:evaluation,
        :english,
        evaluated: user1,
        evaluator: user2,
        evaluation_date: 3.week.ago,
        created_at: 5.month.ago,
        company: admin_user.company
      )
    end
```
You should use `do ... end` instead of `{}` when using a multi line block :nail_care: 

I'd say you could also make the tabs just expand once inside the `create` methods instead of aligning them with `:evaluation`

spec/features/admin/evaluations_spec.rb
```diff
@@ -10,38 +10,110 @@
...
+    let!(:user2) { create(:user,
+                          active: true,
+                          company: admin_user.company) }
+
+    let!(:evaluation1) { create(:evaluation,
+                                :english,
+                                evaluated: user1,
+                                evaluator: user2,
+                                evaluation_date: 3.week.ago,
+                                created_at: 5.month.ago,
```
rhian-cs, line 26:
```suggestion
                                evaluation_date: 3.weeks.ago,
                                created_at: 5.months.ago,
```

spec/features/admin/evaluations_spec.rb
```diff
@@ -10,38 +10,110 @@
...
+                                company: admin_user.company) }
+
+    let!(:evaluation2) { create(:evaluation,
+                                :performance,
+                                evaluator: user1,
+                                evaluated: user2,
+                                evaluation_date: 5.week.ago,
+                                created_at: 8.month.ago,
+                                company: admin_user.company) }
```
rhian-cs, line 35: The previous points also apply here.

spec/features/admin/evaluations_spec.rb
alexandresaura: Thanks, I did it, and I didn't know about the differences between both lets, I will search for now.

spec/features/admin/evaluations_spec.rb
```diff
@@ -10,38 +10,110 @@
   end
 
   describe 'Filters' do
+    let!(:user1) { create(:user,
+                          active: true,
```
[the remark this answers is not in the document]
alexandresaura: Thanks for this tip, I'm not familiar with factories

spec/features/admin/evaluations_spec.rb
alexandresaura: Good catch!

spec/features/admin/evaluations_spec.rb
alexandresaura: I used some example in the code that was using multi-line with `{}` and the editor was saying to use `do ... end` but I didn't know if was more precise and I leave it. Thanks for the suggestion and about the indentation, the editor warn about it:
image

And if I do this way It works:
image

But I prefer the first one, because is simple to search locations that are using `create(:evaluation` this way.

rhian-cs (approved):

Issue #105, opened 2022-04-20
Create an endpoint to show a punch on API


alexandresaura:
We need to create an endpoint that will show all information of a authenticated user's punch.

Thekote:
Hey! Can I try this one?

PR #121, opened 2022-05-26
Show one Punch endpoint to authenticated user

Thekote:
Issue: #105 
This PR adds an endpoint to list all punches registered to the authenticated user.

This is how the response is working:
```json
{
    "created_at": "2022-04-26T09:53:22.722-03:00",
    "from": "2021-10-26T13:00:00.000-03:00",
    "to": "2021-10-26T16:00:00.000-03:00",
    "delta_as_hour": "03:00",
    "extra_hour": false,
    "project": {
        "name": "Punchclock"
    }
}

```

### DIFF
--- a/spec/support/controller_helpers.rb
+++ b/spec/support/controller_helpers.rb
@@ -4,8 +4,6 @@
 # Example:
 #   let(:headers) { { token: user.token } }
 RSpec.shared_examples 'an authenticated resource action' do
-  it { is_expected.to have_http_status(:ok) }
-
   it 'returns content type json' do
     expect(subject.content_type).to eq 'application/json; charset=utf-8'
   end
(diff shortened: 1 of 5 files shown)


spec/support/controller_helpers.rb
```diff
@@ -4,8 +4,6 @@
 # Example:
 #   let(:headers) { { token: user.token } }
 RSpec.shared_examples 'an authenticated resource action' do
-  it { is_expected.to have_http_status(:ok) }
```
fabiosammy, line 7: why?

Thekote: I removed this line because each endpoint will have a different http status response.

adbatista (approved):

adbatista (approved):

Issue #123, opened 2022-06-06
The "approve in batch" does not save "approved_at" and "reviewed_by" information


wenderjean:
Navigating to "Contributions" page, you're able to select a bunch of items in the list and approve/reject them by using the feature "batch actions". There is an issue with this feature because it doesn't fill the information related to the action itself, I mean, "approved_at" and "reviewed_by".

Relates to:
  - https://github.com/Codeminer42/Punchclock/issues/109

![image](https://user-images.githubusercontent.com/941776/172168178-687d60f6-5982-4c5a-887b-3bf66ffe8ba7.png)

PR #124, opened 2022-06-07
Fix contributions batch actions

andreLumor:
Related issue: #123 

This PR fixes the contributions batch approve and reject.

Observed Problem:
Every review resulted in three database calls, one on approve or reject action, and then two called by SaveContributionReviewer service object.

Proposed changes:
Remove SaveContributionReviewer service object.
Contribution's approve and reject now receive the reviewer's id as argument.
Create a proc on contribution approve and reject that updates reviewer info.


Gemfile
```diff
@@ -10,6 +10,7 @@ gem 'rails', '~> 6.1.5.1'
 
 gem 'bootsnap', require: false
 
+gem 'rack', '~> 2.2.3.1'
```
adbatista, line 13: you don't need this

andreLumor: This update was due to CI security

adbatista: is a dependency of rails, if you just do `bundle update rack` will update the gemfile.lock and you don't need to add the gem here

andreLumor: Oh, good to know!

It is already fixed.

PR #125, opened 2022-06-08
Fix delta calculation

joaoGabriel55:


LucasCarrias:
Great job!

app/javascript/packs/components/Form.js
```diff
@@ -141,6 +160,29 @@ class Form extends React.Component{
     this.setState({ hasConfirmedOperation: true, selectedProject: "" })
   }
 
+  areValidWorkPeriods(from1, to1, from2, to2) {
+    if (compareHours({ firstHour: from1, secondHour: to1 }) && compareHours({ firstHour: from2, secondHour: to2 }))
+      return true;
+    if (compareHours({ firstHour: from1, secondHour: from2 }) && compareHours({ firstHour: to1, secondHour: to2 }))
+      return true;
```
abarbosa-godaddy, line 167: Could it be more clear making something like is shown below?

```suggestion
    const cond1 = compareHours({ firstHour: from1, secondHour: to1 }) && compareHours({ firstHour: from2, secondHour: to2 });
   const cond2 = compareHours({ firstHour: from1, secondHour: from2 }) && compareHours({ firstHour: to1, secondHour: to2 });
   return cond1 || cond2;
```

NOTE: You can define more meaningful names instead of `cond1` and `cond2`.

joaoGabriel55: Very nice tip @abarbosa-godaddy!!! I will improve this. Thanks

app/controllers/dashboard_controller.rb
```diff
@@ -10,21 +10,44 @@ def sheets
   end
 
   def save
-    deletes = Array.wrap(params['delete']).concat(params['add'].keys)
+    return head :bad_request if invalid_day_periods?
```
wenderjean, line 13: `:unprocessable_entity` sounds more appropriate to me considering we're dealing with a kind of param validation. Another thing is about being clear on which error we're dealing with (returning a clear message).

app/controllers/dashboard_controller.rb
```diff
@@ -10,21 +10,44 @@ def sheets
...
+
+    deletes = Array.wrap(punch_params['delete']).concat(params_add.keys)
```
wenderjean, line 15: What do you think about moving this entire logic for this `create` method into a service object? Something like `CreatePunchesInBatch.call(additions: [], deletions: [])`

app/controllers/dashboard_controller.rb
```diff
@@ -10,21 +10,44 @@ def sheets
...
+
   def bulk_params(param)
     param.values.flatten.map{|p| p.slice :from, :to, :project_id }
   end
+
+  def invalid_day_periods?
+    params_add.to_h.one? { |_, periods| invalid_period?(periods) }
+  end
+
+  def invalid_period?(periods)
```
wenderjean, line 45: It seems to me this validation could be done like this: 
```ruby
periods.map do |p|
  attributes = p.slice(:from, :to).transform_values(&DateTime.method(:parse))
  
  Punch.new(attributes)
end.reject(&:valid?).any?
```

app/controllers/dashboard_controller.rb
joaoGabriel55: Yeah, very good tip! Thanks!

app/controllers/dashboard_controller.rb
joaoGabriel55: Yeah, would be great!

joaoGabriel55:
Hi @wenderjean, can you take a look at this last commit?

adbatista:
specs breaking

joaoGabriel55:
> specs breaking

OMG. I will take a look

PR #126, opened 2022-07-04
Upgrade Rails to v7

wenderjean:
Must have for deploy

- [ ] Rename env `OTP_ENCRYPTION_KEY` to `OTP_SECRET_ENCRYPTION_KEY`
- [ ] Regenerate credentials for staging/production


adbatista:
To upgrade ruby we need to find an alternative for the gem "validates_timeliness" that is not compatible with rails 3 yet

Issue #130, opened 2022-07-18
Add filter for "reviewed_at" attribute in admin page


wenderjean:
I'd like to filter approved contributions using the new attribute "reviewed_at".

Plus:
1. Remove the column "Empresa" in the contributions table please.
2. Remove filter "Estado" considering we're already using scopes at the top the contributions list

![image](https://user-images.githubusercontent.com/941776/179509503-a10e82b5-66d1-4420-ad9c-d044d992c253.png)

PR #133, opened 2022-07-19
Feat/reviewed at filter

andreLumor:
Resolves: #130 

What was done:
Fixed contributions seeds that broke on #124 (now seeded contributions as self reviewed)
Removed state filter
Removed Company from table

Updated Filters:
image

### DIFF
--- a/app/admin/contributions.rb
+++ b/app/admin/contributions.rb
@@ -6,7 +6,7 @@ ActiveAdmin.register Contribution do
 
   filter :company, as: :select, if: proc { current_user.super_admin? }
   filter :user, as: :select, collection: proc { CompanyUsersCollectionQuery.new(current_user).call }
-  filter :state, as: :select, collection: Contribution.aasm.states_for_select
+  filter :reviewed_at
   filter :created_at
 
   member_action :approve, method: :put, only: :index
@@ -41,7 +41,9 @@ ActiveAdmin.register Contribution do
   index download_links: [:xls, :text] do
     selectable_column
     column :user
-    column :company
+    if current_user.super_admin?
+      column :company
+    end
     column :link do |contribution|
       link_to contribution.link, contribution.link, target: :blank
     end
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -122,7 +122,7 @@ def create_user_contribution(user:, repository:, date:)
       contrib.company_id = user.company.id
       contrib.repository = repository
       contrib.created_at = date
-      contrib.approve
+      contrib.approve(user.id)
```
andreLumor, line 125: Another option would be leaving seeded contributions at their initial state, but I was concerned about breaking tests.

app/admin/contributions.rb
```diff
@@ -6,7 +6,7 @@
 
   filter :company, as: :select, if: proc { current_user.super_admin? }
   filter :user, as: :select, collection: proc { CompanyUsersCollectionQuery.new(current_user).call }
-  filter :state, as: :select, collection: Contribution.aasm.states_for_select
```
adbatista, line 9: why did you remove this?

db/seeds.rb
adbatista: the seeds file is not used in the tests

app/admin/contributions.rb
andreLumor: Also removed as asked on the issue #130

db/seeds.rb
andreLumor: Oh, ok. I'll remove this to a fix PR

adbatista: there's no need, is just to let you know that the seed are not loaded in the tests

app/admin/contributions.rb
```diff
@@ -41,7 +41,6 @@
   index download_links: [:xls, :text] do
     selectable_column
     column :user
-    column :company
```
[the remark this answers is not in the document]
adbatista: instead of removing, change do display this column only is the current_user is superadmin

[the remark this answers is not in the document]
andreLumor: done!

adbatista (approved):

PR #134, opened 2022-07-19
Feat/api jwt auth

andreLumor:
Resolves: #102

What was done:
Created service objects to handle JWT generate, validate and renew
Created token request endpoint
Created token renew endpoint
Set up headers authorization middleware for API endpoints

Most of the implementation was done while pairing with @halan and @Thekote


app/controllers/api/v1/api_controller.rb
```diff
@@ -1,21 +1,22 @@
...
-        return render json: { message: 'Missing Token' }, status: :unauthorized if token.nil?
-        return render json: { message: 'Invalid Token' }, status: :unauthorized unless User.exists?(token: token)
+        rescue Jwt::Verify::InvalidError
+          render json: { error: :unauthorized }, status: :unauthorized
       end
 
-      def token
-        @token ||= request.headers[:token]
+      def bearer_token
+        request.headers['Authorization'].match(/^Bearer (\S+)$/)[1]
```
hpinheiro-godaddy, line 15: you can memoize this

app/controllers/api/v1/token_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+        @token = Jwt::Renew(bearer_token).generate!
+        render json: token_as_json, status: :created
+      end
+
+      private
+      def token_as_json
+        { access_token: @token.to_s }
+      end
+
+      def bearer_token
```
hpinheiro-godaddy, line 29: you have this inherited from `ApiController`

app/controllers/api/v1/token_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      # POST /api/v1/token/request
+      def request
+        @user = User.find_by(user_params.slice(:email))
+
+        if @user && @user.valid_password?(user_params[:password])
+          @token = Jwt::Generate.new(user_id: @user['id'])
+          render json: token_as_json, status: :created
+        else
+          render json: { error: "Usuário ou Senha incorretos" }, status: :unprocessable_entity
```
hpinheiro-godaddy, line 14: i18n?

app/services/jwt/generate.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def token
+    JWT.encode(@payload, ENV['JWT_SECRET_KEY'])
+  end
+
+  alias_method :to_s, :token
+
+  private
+  def generate_next_jti(last_jti)
+    # toDo: define jti blacklist claim
```
hpinheiro-godaddy, line 19: I think that this toDo is outdated.

app/services/jwt/verify.rb
```diff
@@ -0,0 +1,19 @@
...
+  def self.call(token)
+    new(token).call
+  end
+
+  def initialize(token)
+    @token = token[:token]
+  end
+
+  def call
+    # toDo: define jti blacklist claim
```
hpinheiro-godaddy, line 12: that toDo is outdated

db/schema.rb
```diff
@@ -10,7 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_06_17_142319) do
+ActiveRecord::Schema.define(version: 2022_06_17_142319) do
```
hpinheiro-godaddy, line 13: You don't have migrations on that PR, so, don't send schema changes.

app/controllers/api/v1/api_controller.rb
```diff
@@ -1,21 +1,22 @@
...
+      end
+
+      def current_user
+        @current_user = User.find(@current_user_id)
```
hpinheiro-godaddy, line 19: What happens if I call it without calling `require_valid_jwt!` first?

hpinheiro-godaddy (changes requested):
Specs missing.

andreLumor:
Feature implemented in pair with @hpinheiro-godaddy

spec/services/jwt/generate_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    it 'returns correct parse to json' do
+      expect(subject.to_json).to eq({ access_token: subject.token }.to_json)
+    end
+  end
+
+  describe '.token' do
+    subject { class_instance.token }
+    let(:class_instance) { described_class.new(user_id: 1) }
```
andreLumor, line 31: Had to save class instance since SecureRandom is not mocked

adbatista: this should be the subject

spec/services/jwt/generate_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(subject.to_h).to eq({ access_token: subject.token })
+    end
+
+    it 'returns correct parse to json' do
+      expect(subject.to_json).to eq({ access_token: subject.token }.to_json)
+    end
+  end
+
+  describe '.token' do
+    subject { class_instance.token }
```
adbatista, line 30: don't add the exercise step in the subject

spec/services/jwt/generate_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    let(:class_instance) { described_class.new(user_id: 1) }
+    let(:payload) { class_instance.payload }
+    let(:decode_token) { JWT.decode(subject, ENV['JWT_SECRET_KEY']) }
+    
+    it 'calls JWT::encode' do
+      allow(JWT).to receive(:encode).and_return('expected_token')
+      expect(subject).to eq('expected_token')
+      expect(JWT).to have_received(:encode).with(payload, ENV['JWT_SECRET_KEY'])
```
adbatista, line 38: add only on expect per test

spec/support/controller_helpers.rb
```diff
@@ -33,6 +33,6 @@
   end
 
   it 'returns a missing token message' do
```
adbatista, line 35: the title of spec does not match with the spec now

.env.example
```diff
@@ -7,3 +7,4 @@ BULLET_ADD_FOOTER=false
 GITHUB_OAUTH_TOKEN=
 OTP_SECRET_ENCRYPTION_KEY=1ex1ex1ex1ex1ex1ex1ex1ex1ex1ex1ex1ex1ex1ex1ex1ex1e
 NEWS_EMAIL=news@cm42.io
+JWT_SECRET_KEY=
```
hpinheiro-godaddy, line 10: it can work empty like this?

app/controllers/api/v1/token_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+        if @user && @user.valid_password?(params[:password])
+          render json: Jwt::Generate.new(user_id: @user.id), status: :created
+        else
+          render json: { error: I18n.t('errors.messages.wrong_credentials') }, status: :unprocessable_entity
+        end
+      end
+
+      # POST /api/v1/token/refresh
+      def refresh_token
+        render json: Jwt::Renew.new(bearer_token).generate!.to_json, status: :created
```
hpinheiro-godaddy, line 19: `.to_json` isn't necessary here. You can use the same strategy that you did on `#request_token`

hpinheiro-godaddy (changes requested):
just a small change

.env.example
andreLumor: Yeah, it uses an empty string

hpinheiro-godaddy (dismissed):
🚢

spec/controllers/api/v1/token_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'spec_helper'
+require  "#{Rails.root}/spec/support/controller_helpers.rb"
```
adbatista, line 2: rspec already has a config to load files under spec/support, check the spec_helper it might be commented

spec/controllers/api/v1/token_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+describe Api::V1::TokenController, type: :controller do
+  let(:user) { create(:user) }
+  describe 'POST api/v1/request' do
+    subject { post :request_token, params: params }
```
adbatista, line 7: please don't put the exercise test step in the subject or any setup phase

adbatista, line 7: please don't put the exercise test step in the subject or any setup phase

adbatista (changes requested):

spec/controllers/api/v1/token_controller_spec.rb
andreLumor: I executed the endpoint individually in every test, as in the 'contributions_controller.spec'. Is this the correct way?
ex:
```rb
it 'returns created status' do
  post :request_token, params: params
  expect(response).to have_http_status(:created)
end
```

adbatista: yep, tests are the case when you sacrifice some DRY to keep more readable, so is better to look at the test and know what is being exercised, that's why subject should be only setup too

spec/services/jwt/renew_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    Rails.cache.clear
+  end
+
+  context '.revoke!' do
+    subject { described_class.new(token).revoke! }
+
+    it 'adds token to jti blacklist' do
+      allow(Rails.cache).to receive(:write)
+      subject
+      expect(Rails.cache).to have_received(:write).with("jwt_jti_denied/#{jti}", true)
```
adbatista, line 19: if you read only this spec, can you tell what is being tested, is `subject` descriptive enough?

spec/controllers/api/v1/token_controller_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+
+      it 'invalidates old access token' do
+        old_token = current_jwt
+        post :refresh_token
+        expect { Jwt::Verify.call(old_token) }.to raise_error(Jwt::Verify::InvalidError)
+      end
+    end
+
+    context 'when Authentication header is invalid' do
```
hpinheiro-godaddy, line 77: It could be important to spec a case using `alg:none` (described here https://github.com/jwt/ruby-jwt#none).

Currently, the default setting is `alg:HS256` (here https://github.com/jwt/ruby-jwt/blob/master/lib/jwt/configuration/decode_configuration.rb#L26) but could be a good idea to ensure it on your specs.

The plan is to generate an `alg:none` JWT and request a protected endpoint. It should be unauthorized.

hpinheiro-godaddy (changes requested):

spec/controllers/api/v1/token_controller_spec.rb
andreLumor: Done!

adbatista (approved):

PR #141, opened 2022-07-26
Fix/several fixes

Caps-Looking:
Fix scope all to show only active users
Fix user edit to show date correctly
Updates query to handle finished allocations as unallocated
Changes spreadsheet tab name
Changes spreadsheet columns

### DIFF
--- a/app/decorators/user_decorator.rb
+++ b/app/decorators/user_decorator.rb
@@ -38,4 +38,8 @@ class UserDecorator < Draper::Decorator
   def overall_score
     model.overall_score || I18n.t('not_evaluated')
   end
+
+  def office_head_name
+    model.office.head&.name || 'N/A'
+  end
 end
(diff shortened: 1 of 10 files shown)


app/decorators/user_decorator.rb
```diff
@@ -38,4 +38,8 @@ def performance_score
   def overall_score
     model.overall_score || I18n.t('not_evaluated')
   end
+
+  def office_head_name
```
wenderjean, line 42: Can we add a test for this in `user_decorator_spec.rb`?

spec/spreadsheets/allocations_spreadsheet_spec.rb
```diff
@@ -9,23 +9,19 @@
...
-      Allocation.human_attribute_name('start_at'),
-      Allocation.human_attribute_name('end_at'),
-      Allocation.human_attribute_name('created_at'),
-      Allocation.human_attribute_name('updated_at'),
+      Allocation.human_attribute_name('end_at')
     ]
   end
 
-  let (:body_attributes) do
+  let(:body_attributes) do
```
wenderjean, line 18: We should use explicit values in this let, I mean, you should test the end result for the office's head name instead of decorating the user and taking these attributes.

wenderjean (approved):

PR #142, opened 2022-07-29
Fix/allocation query

Caps-Looking:
Fix allocations query order.
Fix date bug in allocation spreadsheet test caused by fixed date.

### DIFF
--- a/app/queries/allocations_and_unalocated_users_query.rb
+++ b/app/queries/allocations_and_unalocated_users_query.rb
@@ -17,7 +17,7 @@ class AllocationsAndUnalocatedUsersQuery
         (allocations.end_at > NOW() OR allocations.end_at IS NULL)'
       )
       .where(where)
-      .order('start_at NULLS LAST, end_at NULLS LAST, users.name')
+      .order('end_at NULLS LAST, start_at NULLS LAST, users.name')
   end
 
   private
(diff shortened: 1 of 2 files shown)


spec/spreadsheets/allocations_spreadsheet_spec.rb
```diff
@@ -26,7 +26,7 @@
       'backend',
       'Punchclock',
       'Wender',
-      I18n.l(Date.parse('27/07/2022'), format: :long)
+      I18n.l(Date.today + 10.days, format: :long)
```
wenderjean, line 29: You can use `10.days.from_now`

app/queries/allocations_and_unalocated_users_query.rb
```diff
@@ -17,7 +17,7 @@ def call
         (allocations.end_at > NOW() OR allocations.end_at IS NULL)'
       )
       .where(where)
-      .order('start_at NULLS LAST, end_at NULLS LAST, users.name')
+      .order('end_at NULLS LAST, start_at NULLS LAST, users.name')
```
wenderjean, line 20: Can we get rid of this `start_at` ordering?

Caps-Looking: No. If we remove the `start_at` ordering, the allocations without end will appear mixed with the unallocated ones.
image

wenderjean (approved):

PR #145, opened 2022-08-02
add dev-caching tmp file

andreLumor:
Since /tmp files are ignored, def-caching file had to be force added.
Now development environment is caching correctly.


adbatista:
why do we need this?

andreLumor:
> why do we need this?

Rails uses this file as cache in development environment.
Without it we are unable to deny JWT tokens (jti denylist is cached).

adbatista:
but i still don't get why this need to be added on git, if the file is being created when running, should be fine

andreLumor:
The file is not created when running. If it doesn't exist, rails simply won't do caching on development env.

adbatista:
I got it, but if this is always enabled we may have some issues in some other parts of the app, is better to enable it only if needed
running `rails dev:cache`

Issue #151, opened 2022-08-09
Create a new page to list all engineers and their allocation


adbatista:
the menu item should be under the Users menu

EN: allocation chart
Pt-BR: organograma

the users list should have the columns:
name -> engineer first and last name with a link to the engineer show page
skills -> engineer skills
client -> current engineer allocation (if has)
end -> allocation end if has, show nothing if nil

the order should be allocated with end date first, allocated without end date and unallocated
there's already a class to return the data ordered

PR #156, opened 2022-08-11
Adds allocation chart page

mateuscelio:
Adds page as specified in #151

image

image

### DIFF
--- a/app/admin/allocation_chart.rb
+++ b/app/admin/allocation_chart.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+ActiveAdmin.register_page 'Allocation Chart' do
+  menu label: proc { I18n.t('allocation_chart') }, parent: User.model_name.human(count: 2)
+
+  content title: I18n.t('allocation_chart') do
+    panel I18n.t('allocation_chart') do
+      allocations = AllocationsAndUnalocatedUsersQuery.new(Allocation, current_user.company).call
+      paginated_collection(allocations.page(params[:page]), download_links: false) do
+        table_for collection do
+          column(I18n.t('name'), :user)
+          column(I18n.t('skills')) { |allocation| allocation.user.skills.map(&:title).join(', ') }
+          column(I18n.t('client'), :project)
+          column(I18n.t('allocated_until'), :end_at)
+        end
+      end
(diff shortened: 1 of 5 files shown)


app/admin/allocation_chart.rb
```diff
@@ -0,0 +1,19 @@
...
+
+ActiveAdmin.register_page 'Allocation Chart' do
+  menu label: proc { I18n.t('allocation_chart') }, parent: User.model_name.human(count: 2)
+
+  content title: I18n.t('allocation_chart') do
+    panel I18n.t('allocation_chart') do
+      allocations = AllocationsAndUnalocatedUsersQuery.new(Allocation, current_user.company).call
+      paginated_collection(allocations.page(params[:page]), download_links: false) do
+        table_for collection do
+          column(I18n.t('name'), :user)
```
GabrielRMuller, line 11: This column should also link to the engineer show page, right?

mateuscelio: Passing an ActiveRecord instance to `column` method automatically creates a link. Added a screenshot.

spec/features/admin/allocation_chart_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  describe 'Index' do
+    let(:project) { create(:project, company: company) }
+    let!(:user) { create(:user, allocations: [build(:allocation, project: project, company: company)], company: company) }
+
+    before { visit '/admin/allocation_chart' }
+
+    it 'have table with fields "Nome", "Habilidades", "Cliente", "Alocado até"' do
+      within 'table' do
+        expect(page).to have_text('Nome') & have_text('Habilidades') & have_text('Cliente') & have_text('Alocado até')
```
GabrielRMuller, line 21: I would suggest adding `aggregate_failures` in here, so in case one of these fields fails we know exactly which one is it. As of now if one of these fields fails the spec will throw a single error without specifying which one is it. https://relishapp.com/rspec/rspec-core/docs/expectation-framework-integration/aggregating-failures

adbatista: composable matches already do this

GabrielRMuller (approved):

Issue #152, opened 2022-08-09
Highlight allocation chart rows based on the end date


adbatista:
unallocated engineers: background light green
end date in the past: red
end date < 30 days: light red
end date < 60 days yellow

PR #158, opened 2022-08-15
Highlight allocation chart rows based on the end date

mateuscelio:
#152

image

### DIFF
--- a/app/admin/allocation_chart.rb
+++ b/app/admin/allocation_chart.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+include ActiveAdmin::AllocationChartHelper
+
 ActiveAdmin.register_page 'Allocation Chart' do
   menu label: proc { I18n.t('allocation_chart') }, parent: User.model_name.human(count: 2)
 
@@ -11,9 +13,33 @@ ActiveAdmin.register_page 'Allocation Chart' do
           column(I18n.t('name'), :user)
           column(I18n.t('skills')) { |allocation| allocation.user.skills.map(&:title).join(', ') }
           column(I18n.t('client'), :project)
-          column(I18n.t('allocated_until'), :end_at)
+          column(I18n.t('allocated_until'), class: 'allocated-column') do |allocation|
+            build_allocation_status_cell(allocation)
+          end
         end
       end
     end
+  end
+end
+
+def build_allocation_status_cell(last_allocation)
+  allocation_status = ActiveAdmin::AllocationChartHelper.allocation_status_for(last_allocation: last_allocation)
+
+  unless last_allocation.end_at.nil?
+    return column_cell_base(content: last_allocation.end_at, allocation_status: allocation_status)
+  end
+
+  content = if allocation_status == ActiveAdmin::AllocationChartHelper::Status::NOT_ALLOCATED
+              I18n.t('not_allocated')
+            else
+              I18n.t('allocated')
+            end
+
+  column_cell_base(content: content, allocation_status: allocation_status)
+end
+
+def column_cell_base(content:, allocation_status:)
+  div class: "allocated-column__status  allocated-column__status--#{allocation_status}" do
+    content
   end
 end
--- a/app/admin/allocation_chart.rb
+++ b/app/admin/allocation_chart.rb
@@ -23,19 +23,25 @@ ActiveAdmin.register_page 'Allocation Chart' do
 end
 
 def build_allocation_status_cell(last_allocation)
-  allocation_status = ActiveAdmin::AllocationChartHelper.allocation_status_for(last_allocation: last_allocation)
+  allocation_status = allocation_status_for(last_allocation: last_allocation)
 
-  unless last_allocation.end_at.nil?
-    return column_cell_base(content: last_allocation.end_at, allocation_status: allocation_status)
-  end
+  column_cell_base(
+    content: allocation_status_content(
+      last_allocation: last_allocation,
+      allocation_status: allocation_status
+    ),
+    allocation_status: allocation_status
+  )
+end
 
-  content = if allocation_status == ActiveAdmin::AllocationChartHelper::Status::NOT_ALLOCATED
-              I18n.t('not_allocated')
-            else
-              I18n.t('allocated')
-            end
+def allocation_status_content(last_allocation:, allocation_status:)
+  return last_allocation.end_at if last_allocation.end_at
 
-  column_cell_base(content: content, allocation_status: allocation_status)
+  unallocated?(allocation_status) ? I18n.t('not_allocated') : I18n.t('allocated')
+end
+
+def unallocated?(allocation_status)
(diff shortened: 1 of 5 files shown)


app/admin/allocation_chart.rb
```diff
@@ -11,9 +13,33 @@
...
         end
       end
     end
   end
 end
+
+def build_allocation_status_cell(last_allocation)
+  allocation_status = ActiveAdmin::AllocationChartHelper.allocation_status_for(last_allocation: last_allocation)
+
+  unless last_allocation.end_at.nil?
```
GabrielRMuller, line 28: nothing wrong here! but I would suggest avoiding using the inverse logic of `unless`, since it tends to be a bit harder to read because we're more used to `if`s. So what do you think of changing this to something like:
```ruby
if last_allocation.end_at
  return column_cell_base(content: last_allocation.end_at, allocation_status: allocation_status)
end
```

app/admin/allocation_chart.rb
```diff
@@ -11,9 +13,33 @@
...
+    return column_cell_base(content: last_allocation.end_at, allocation_status: allocation_status)
+  end
+
+  content = if allocation_status == ActiveAdmin::AllocationChartHelper::Status::NOT_ALLOCATED
```
GabrielRMuller, line 32: for the sake of making this method smaller, is it possible to move this if/else into a separate method? then you'd call it in here, like so:
```ruby
content = allocation_status_content
```

you could even pass the method directly as an argument in the `content` key in line 38:
```ruby
column_cell_base(content: allocation_status_content, allocation_status: allocation_status)
```
(the name is just a suggestion, feel free to name something more appropriate)

app/assets/stylesheets/active_admin.scss
```diff
@@ -73,4 +73,31 @@ $active-admin-a-border: #c7c7c7;
...
+
+  &__status {
+    padding: 5px;
+
+    &--exp-in-60-days {
+      background-color: #ffec3d;
+      color: black;
+    }
+    &--exp-in-30-days {
+	    background-color: #ff7875;
```
GabrielRMuller, line 90: extra tab in here and in line 94 :bomb:

app/helpers/active_admin/allocation_chart_helper.rb
```diff
@@ -0,0 +1,35 @@
...
+    class Status
+      EXP_IN_60_DAYS = 'exp-in-60-days'
+      EXP_IN_30_DAYS = 'exp-in-30-days'
+      EXPIRED = 'expired'
+      NOT_ALLOCATED = 'not-allocated'
+      ALLOCATED = 'allocated'
+    end
+
+    def allocation_status_for(last_allocation:)
+      return Status::NOT_ALLOCATED if last_allocation&.id.nil?
```
GabrielRMuller, line 14: is it necessary to check the `id` explicitly?

app/helpers/active_admin/allocation_chart_helper.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      end_date = last_allocation&.end_at
+      return Status::ALLOCATED if end_date.nil?
+
+      day_difference = (end_date - Date.today unless end_date.nil?)
```
GabrielRMuller, line 19: maybe it's not necessary to check if the `end_date` is `nil`, since we're doing this on line 17

GabrielRMuller (commented):
nice! left a few suggestions for now

app/helpers/active_admin/allocation_chart_helper.rb
mateuscelio: Yes because some `last_allocation` values can be "empty" `Allocation` instances

app/admin/allocation_chart.rb
```diff
@@ -11,9 +13,36 @@
...
+  allocation_status = ActiveAdmin::AllocationChartHelper.allocation_status_for(last_allocation: last_allocation)
+
+  column_cell_base(
+    content: allocation_status_content(last_allocation: last_allocation,
+                                       allocation_status: allocation_status), allocation_status: allocation_status
+  )
+end
+
+def allocation_status_content(last_allocation:, allocation_status:)
+  return last_allocation.end_at if last_allocation.end_at
```
GabrielRMuller, line 35: I feel like this is something that might be useful as a method on the `Allocation` model. That way, the method name will describe what we want when checking if the `end_at` field exists and will make the code more readable. So how about we add this in `app/models/allocation.rb`?

```ruby
def currently_allocated?
  last_allocation.end_at?
end
```

Then the code in here would be:
```ruby
return last_allocation.end_at if last_allocation.currently_allocated?
```

app/admin/allocation_chart.rb
```diff
@@ -11,9 +13,36 @@
...
+
+  if allocation_status == ActiveAdmin::AllocationChartHelper::Status::NOT_ALLOCATED
```
GabrielRMuller, line 37: I think we can make this if/else statement a bit cleaner!

How about passing this check into a smaller private method and turning this into a ternary check? like so:
```ruby
if unallocated?(allocation_status) ? I18n.t('not_allocated') : I18n.t('allocated')
...
private def unallocated?(status)
  status == ActiveAdmin::AllocationChartHelper::Status::NOT_ALLOCATED
end
```
That way we improve the readabilty of this `if` statement and we reduce it to a single line

app/admin/allocation_chart.rb
```diff
@@ -11,9 +13,36 @@
...
+            build_allocation_status_cell(allocation)
+          end
         end
       end
     end
   end
 end
+
+def build_allocation_status_cell(last_allocation)
+  allocation_status = ActiveAdmin::AllocationChartHelper.allocation_status_for(last_allocation: last_allocation)
```
GabrielRMuller, line 26: by the way, do we need to call `ActiveAdmin::AllocationChartHelper` in here if we're including it on line 3?

app/admin/allocation_chart.rb
mateuscelio: An allocation does not need and `end_at` date. So I belive that this method can lead to confusion. What about this method simplification?

```rb
def allocation_status_content(last_allocation:, allocation_status:)
      last_allocation.end_at || unallocated?(allocation_status) ? I18n.t('not_allocated') : I18n.t('allocated')
end
```

GabrielRMuller: > An allocation does not necessarily needs an end_at date

Ok, the method name was just an idea, what is it that we're checking for exactly then? when does an allocation have an end_date?

mateuscelio: Deciding if the column content is `end_at` date when it exists or a text based on allocation status. `end_date` is setting in allocation creation/update.

GabrielRMuller: Ok, so I'd say we leave as is

GabrielRMuller (approved):

PR #159, opened 2022-08-15
Adds columns level and specialty to allocation chart table

mateuscelio:
image

### DIFF
--- a/config/locales/en/en.yml
+++ b/config/locales/en/en.yml
@@ -60,3 +60,4 @@ en:
   skills: 'Skills'
   client: 'Client'
   allocated_until: 'Allocated until'
+  specialty: "Specialty"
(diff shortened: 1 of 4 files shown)


config/locales/en/en.yml
```diff
@@ -60,3 +60,4 @@ en:
   skills: 'Skills'
   client: 'Client'
   allocated_until: 'Allocated until'
+  specialty: "Especialty"
```
GabrielRMuller, line 63:
```suggestion
  specialty: "Specialty"
```

GabrielRMuller (changes requested):
Looks good! but can you make it so this pr only has one commit?

GabrielRMuller (approved):

Issue #163, opened 2022-08-16
Add `ongoing` property to `Allocation` model


mateuscelio:
Add `ongoing` property to `Allocation` model.
This property must be a `boolean` with `true` as default but will temporarily be default to `false`.
An engineer can only have one allocation at a time with this property equals to `true`.

PR #164, opened 2022-08-18
Add ongoing property to Allocation model

mateuscelio:
#163

* Add `ongoing` property to Allocation model
* Update `allocations` active admin pages to present/manipulate ongoing info
* Update `user` active admin page to present ongoing info
* Create task to set current users allocation ongoing property to true

## Screen shots:

* image 1

image

* image 2

image

* image 3

image

* image 4

image

* image 5

image

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -9,26 +9,29 @@ class Allocation < ApplicationRecord
   validates_date :start_at
   validate :end_date
   validate :unique_period, unless: :end_before_start?
+  validates :ongoing, uniqueness: { scope: :user, message: :uniqueness },
+                      if: :ongoing
 
   delegate :office_name, to: :user
 
-  scope :ongoing, -> { 
+  scope :ongoing, lambda {
     where("end_at > :today OR end_at is NULL", today: Date.current)
-    .where(user_id: User.active)
-    .order(start_at: :desc) 
+      .where(user_id: User.active)
+      .order(start_at: :desc)
   }
-  scope :finished, -> {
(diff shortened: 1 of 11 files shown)


GabrielRMuller (changes requested):
Let's separate the linter changes from the feature implementation

lib/tasks/allocations.rake
```diff
@@ -0,0 +1,10 @@
+namespace :allocations do
+  desc 'Update users current allocation ongoing column'
+  task update_users_current_allocation: :environment do
+    ActiveRecord::Base.transaction do
+      User.includes(:allocations).joins(:allocations).merge(Allocation.ongoing).each do |user|
```
GabrielRMuller, line 5: Do we wanna call `includes` before calling `joins` in here?

app/models/allocation.rb
```diff
@@ -9,6 +9,8 @@ class Allocation < ApplicationRecord
   validates_date :start_at
   validate :end_date
   validate :unique_period, unless: :end_before_start?
+  validates :ongoing, uniqueness: { scope: :user, message: :uniqueness },
+                      if: :ongoing
```
GabrielRMuller, line 13: What is the conditional (`if: :ongoing`) checking?

mateuscelio: This will only validate uniqueness for `ongoing` if it has value `true` because a user can only have one allocation with `ongoing` equals to `true`

GabrielRMuller (approved):

GabrielRMuller (approved):

PR #168, opened 2022-08-25
Show one month at a time in multiple punches calendar 

mateuscelio:
image
image

### DIFF
--- a/app/javascript/packs/utils/calendar.js
+++ b/app/javascript/packs/utils/calendar.js
@@ -4,20 +4,21 @@ import moment from 'moment';
 const daysPerWeek = 7;
 const Day = Immutable.Record({day: undefined, inner: undefined, today: false});
 const Week = Immutable.Record({days: Immutable.List()});
+const INCLUSIVITY = '[]' 
 
 export function prev(base){
-  return base.clone().subtract(1, 'M').date(base.date());
+  return base.clone().subtract(1, 'M');
 }
 
 export function next(base){
-  return base.clone().add(1, 'M').date(base.date());
+  return base.clone().add(1, 'M');
 }
 
 export function week(date, range){
   return Immutable.Range(0, daysPerWeek).map((i)=> {
     let day = date.clone().add(i, 'd');
     let [from, to] = range;
-    let inner = day.isBetween(from, to, 'day');
+    let inner = day.isBetween(from, to, 'day', INCLUSIVITY);
     let today = day.isSame(moment(), 'day');
     return new Day({ day: day, inner: inner, today: today });
   });
@@ -31,8 +32,8 @@ export function weeks(start, range){
 }
 
 export function innerRange(base){
-  return [base.clone().subtract(1, 'M').date(base.date()),
-          moment.min(base.clone().add(1, 'd'), moment().add(1, 'd'))];
+  return [base.clone(),
+          moment.min(base.clone().add(1, 'M').subtract(1, 'd'), current())];
 }
 
 export function monthNames(range){
@@ -48,11 +49,11 @@ export function monthNames(range){
 }
 
 export function startDate(base){
-  return base.clone().subtract(1, 'M').date(base.date()).add(1, 'day').day(0);
+  return base.clone().day(0);
 }
 
 export function current() {
-  return moment().add(1, 'M');
+  return moment().utc()
 }
 
 export function constraintMonth(year, month) {
--- a/spec/features/dashboard_spec.rb
+++ b/spec/features/dashboard_spec.rb
@@ -10,11 +10,11 @@ feature 'Punches Dashboard', js: true do
     scenario 'Calendar navigation' do
       visit '/dashboard/2014/12'
 
-      expect(page).to have_content('Nov / Dec 2014')
+      expect(page).to have_content('December 2014')
       find('a', text: '❯').click
-      expect(page).to have_content('Dec 2014 / Jan 2015')
+      expect(page).to have_content('January 2015')
       find('a', text: '❮').click
-      expect(page).to have_content('Nov / Dec 2014')
+      expect(page).to have_content('December 2014')
     end
 
     scenario 'Insert and delete punches' do
@@ -46,15 +46,13 @@ feature 'Punches Dashboard', js: true do
     end
 
     scenario 'Multiple selection through sheets' do
-      visit '/dashboard/2018/02'
+      visit '/dashboard/2018/01'
 
       find('td.inner', text: '19').click
       find('td.inner', text: '20').click
       expect(page).to have_content('Selecionado (2)')
-      expect(page).to have_content('Selecionado (2)')
 
-
-      within 'tbody > tr:nth-child(1)' do
+      within 'tbody > tr:nth-child(3)' do
         expect(page).to have_css("td.selected", text: "19") & have_css("td.selected", text: "20")
       end
 
--- a/test/javascript/packs/utils/calendar.test.js
+++ b/test/javascript/packs/utils/calendar.test.js
@@ -1,4 +1,10 @@
-import { compareHours } from "../../../../app/javascript/packs/utils/calendar";
+import moment from 'moment';
+import { compareHours, prev, next, current, innerRange, startDate } from "../../../../app/javascript/packs/utils/calendar";
+
+const formattedDate = (date) => date.format('YYYY-MM-DD');
+const CURRENT_DATE = '2022-08-15'
+
+
 
 describe("calendar", () => {
   describe("compareHours", () => {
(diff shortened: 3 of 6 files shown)


test/javascript/packs/utils/calendar.test.js
```diff
@@ -1,4 +1,5 @@
-import { compareHours } from "../../../../app/javascript/packs/utils/calendar";
+import moment from 'moment';
+import { compareHours, prev, next, current, innerRange, startDate } from "../../../../app/javascript/packs/utils/calendar";
```
Lukazovic, line 2: `current` is not being used. Shouldn't we have a test for this function?

test/javascript/packs/utils/calendar.test.js
```diff
@@ -24,4 +25,48 @@ describe("calendar", () => {
...
+      const baseDate = moment('2022-06-01').utc()
+      it('returns a range from base date until one day before next month', () => {
+        const range = innerRange(baseDate, currentDate)
+        expect(range[0].format('YYYY-MM-DD')).toBe('2022-06-01')
+        expect(range[1].format('YYYY-MM-DD')).toBe('2022-06-30')
+      });
+    })
+  });
+
+  describe('startDate', () => {
```
Lukazovic, line 66: When we run this test, there is a warning in the console. Shouldn't it be fixed?

image

test/javascript/packs/utils/calendar.test.js
```diff
@@ -24,4 +25,48 @@ describe("calendar", () => {
...
     });
   });
+
+  describe("prev", () => {
+    const base_date = moment('2022-08-01').utc()
+    it('returns previous month in same day', () => {
+      const prev_month = prev(base_date)
+      expect(prev_month.format('YYYY-MM-DD')).toBe('2022-07-01')
+    });
+  });
```
Lukazovic, line 35: In JS we don't use Snake Case:
```suggestion
  describe("prev", () => {
    const baseDate = moment('2022-08-01').utc()
    it('returns previous month in same day', () => {
      const prevMonth = prev(baseDate)
      expect(prevMonth.format('YYYY-MM-DD')).toBe('2022-07-01')
    });
  });
```

test/javascript/packs/utils/calendar.test.js
```diff
@@ -24,4 +25,48 @@ describe("calendar", () => {
...
+
+  describe("next", () => {
+    const base_date = moment('2022-08-01').utc()
+    it('returns next month in same day', () => {
+      const next_month = next(base_date)
+      expect(next_month.format('YYYY-MM-DD')).toBe('2022-09-01')
+    });
+  });
```
Lukazovic, line 43: In JS we don't use Snake Case:
```suggestion
  describe("prev", () => {
    const baseDate = moment('2022-08-01').utc()
    it('returns previous month in same day', () => {
      const prevMonth = prev(baseDate)
      expect(prevMonth.format('YYYY-MM-DD')).toBe('2022-07-01')
    });
  });

  describe("next", () => {
    const baseDate = moment('2022-08-01').utc()
    it('returns next month in same day', () => {
      const nextMonth = next(baseDate)
      expect(nextMonth.format('YYYY-MM-DD')).toBe('2022-09-01')
    });
  });
```

app/javascript/packs/utils/calendar.js
```diff
@@ -6,18 +6,18 @@ const Day = Immutable.Record({day: undefined, inner: undefined, today: false});
...
-  return base.clone().add(1, 'M').date(base.date());
+  return base.clone().add(1, 'M');
 }
 
 export function week(date, range){
   return Immutable.Range(0, daysPerWeek).map((i)=> {
     let day = date.clone().add(i, 'd');
     let [from, to] = range;
-    let inner = day.isBetween(from, to, 'day');
+    let inner = day.isBetween(from, to, 'day', '[]');
```
Lukazovic, line 20: This `'[]` feels weird (with no meaning) I had to search in the Moment docs to understand what it means.
image

What if we can let the meaning of that more explicit? I mean, something like:
```js
const inclusivity = '[]';
// mabe
// const includeBothValues = '[]';
let inner = day.isBetween(from, to, 'day', inclusivity);
```
Would that make sense? Or am I missing something?

app/javascript/packs/utils/calendar.js
```diff
@@ -30,9 +30,9 @@ export function weeks(start, range){
   });
 }
 
-export function innerRange(base){
-  return [base.clone().subtract(1, 'M').date(base.date()),
-          moment.min(base.clone().add(1, 'd'), moment().add(1, 'd'))];
+export function innerRange(base, currentDate=current()){
+  return [base.clone(),
+          moment.min(base.clone().add(1, 'M').subtract(1, 'd'), currentDate)];
```
Lukazovic, line 35: I have never seen a function being called in the argument of another function to determine the default value of an argument. I think you can make it more explicit and readable:
```suggestion
export function innerRange(base, currentDate){
  const validCurrentDate = currentDate || current();

  return [base.clone(),
          moment.min(base.clone().add(1, 'M').subtract(1, 'd'), validCurrentDate)];
```

Lukazovic: Btw, looks like we don't have a test to guarantee we have a default date for `currentDate`. If we remove the default value, all the tests will still pass. Shouldn't we have a test for that?
Peek 2022-08-26 15-17

app/javascript/packs/utils/calendar.js
```diff
@@ -6,18 +6,18 @@ const Day = Immutable.Record({day: undefined, inner: undefined, today: false});
...
-  return base.clone().subtract(1, 'M').date(base.date());
+  return base.clone().subtract(1, 'M');
 }
 
 export function next(base){
-  return base.clone().add(1, 'M').date(base.date());
+  return base.clone().add(1, 'M');
 }
 
 export function week(date, range){
```
Lukazovic, line 16: You modified this function but didn't add tests to guarantee it will always work as expected, should it have tests?

spec/features/dashboard_spec.rb
```diff
@@ -46,15 +46,13 @@
...
+      visit '/dashboard/2018/01'
 
       find('td.inner', text: '19').click
       find('td.inner', text: '20').click
       expect(page).to have_content('Selecionado (2)')
-      expect(page).to have_content('Selecionado (2)')
-
 
-      within 'tbody > tr:nth-child(1)' do
+      within 'tbody > tr:nth-child(3)' do
```
Lukazovic, line 55: Question: why this number has changed? (I don't know the answer, I am just curious about that hahaha)

test/javascript/packs/utils/calendar.test.js
mateuscelio: Test added!

app/javascript/packs/utils/calendar.js
mateuscelio: I added this [test](https://github.com/Codeminer42/Punchclock/pull/168/files#diff-cbdcfd2dcd0b57fe9951a9ec03bdde02a407e0c94435d7c8af73d7a561d8c04dR37). What do you think about it?

spec/features/dashboard_spec.rb
mateuscelio: Before, the 15th day of the month was the base day. So I [changed](https://github.com/Codeminer42/Punchclock/pull/168/files#diff-1358ec85630c58187bcdc2a3760746bab34b1737b015cde67dd1ec2cebd7da52R5) the base day to the 1st. This caused the move of the week lines.

Lukazovic: Got it, thanks!

test/javascript/packs/utils/calendar.test.js
```diff
@@ -24,4 +27,63 @@ describe("calendar", () => {
...
+        const [from, to] = innerRange(baseDate, currentDate)
+        expect(formattedDate(from)).toBe('2022-08-01')
+        expect(formattedDate(to)).toBe('2022-08-15')
+      });
+
+      describe("when called without currentDate", () => {
+        const baseDate = current().subtract(1, 'd')
+
+        it("uses current system date", () => {
+          const [_, to] = innerRange(baseDate);
```
Lukazovic, line 67: We don't need the `_` variable:
```suggestion
          const [, to] = innerRange(baseDate);
```

test/javascript/packs/utils/calendar.test.js
```diff
@@ -24,4 +27,63 @@ describe("calendar", () => {
...
+      });
+    })
+  });
+
+  describe('startDate', () => {
+    it('returns the date of first day of the week relative to base date', () => {
+      const baseDate = moment('2022-07-1', 'YYYY-MM-DD').utc()
+      expect(formattedDate(startDate(baseDate))).toBe('2022-06-26')
+    });
+  });
```
Lukazovic, line 88: I think we can improve the readability of all the tests to make it explicit what each part of it means, check it out what it could look like:
```suggestion
  describe("startDate", () => {
    it("returns the date of first day of the week relative to base date", () => {
      const baseDate = moment("2022-07-1", "YYYY-MM-DD").utc();
      const expectedDate = "2022-06-26";

      const result = startDate(baseDate);
      const formattedResult = formattedDate(result);

      expect(formattedResult).toBe(expectedDate);
    });
  });
```

What do you think about it? Does that make sense?

Lukazovic (changes requested):

app/javascript/packs/utils/calendar.js
Lukazovic: It makes sense to me 👍🏽

test/javascript/packs/utils/calendar.test.js
```diff
@@ -1,4 +1,15 @@
...
+
+const formattedDate = (date) => date.format('YYYY-MM-DD');
+const CURRENT_DATE = '2022-08-15'
+
+beforeAll(() => {
+  jest.useFakeTimers('modern').setSystemTime(new Date(CURRENT_DATE))
+})
+afterAll(() => {
+  jest.useRealTimers()
+})
```
Lukazovic, line 12: Usually, we place these hooks inside the `describe`, so I think we should leave these inside the `describe` but that's not a blocker, so I am going to approve this PR

Lukazovic (approved):

adbatista (approved):

PR #171, opened 2022-08-26
Fix allocation list

adbatista:


spec/features/admin/users_spec.rb
```diff
@@ -17,7 +17,7 @@
     before do
       create(:user)
       create(:office, head: user)
-      create(:allocation, user: allocated_user)
+      create(:allocation, user: allocated_user, ongoing: true)
```
GabrielRMuller, line 20: maybe we should set the `ongoing` value for allocation as true on the factory, feels kinda counter-intuitive to have the field with the default value `true` and its factory as `false` https://github.com/Codeminer42/Punchclock/blob/9a554fc2b428ec6bd0a524dbcc0baa4f86a30883/spec/factories/allocations.rb#L10

GabrielRMuller (approved):

app/models/allocation.rb
```diff
@@ -34,7 +30,11 @@ class Allocation < ApplicationRecord
   def days_until_finish
     return unless end_at
 
-    end_at > Date.current ? (end_at - Date.current).to_i : I18n.t('finished')
+    if ongoing
+      (end_at - Date.current).to_i
```
mateuscelio, line 34: This may result in negative numbers. Is this an expected behavior?

adbatista: yep

PR #172, opened 2022-08-29
Feature/create contributions decorator

### LINKED ISSUE #157: Improvements for "Contributions" page
We have a couple of improvements we'd like to add to the Contributions page:

1. User column, use only the user's first and last name
2. Format creation date using "dd/mm/yyyy"
3. In the reviewer column we should use only the first letter of the first and last name (instead of Wender Freese we should present only WF)
4. Format review date using "dd/mm/yyyy"
5. Remove the default auto-complete behavior from the filters (it overrides the date picker)

[PLUS][BUG]

1. There is a bug during the txt exporting process, it doesn't collect only selected rows, we should fix this

vinas-go:
Resolve #157


app/admin/contributions.rb
```diff
@@ -40,7 +41,9 @@
 
   index download_links: [:xls, :text] do
     selectable_column
-    column :user
+    column :user do |contribution|
```
adbatista, line 44: you can also do `column :user, &:user_short_name`

app/admin/contributions.rb
```diff
@@ -1,23 +1,24 @@
...
-  filter :user, as: :select, collection: proc { CompanyUsersCollectionQuery.new(current_user).call }
+  filter :company, as: :select, if: proc {current_user.super_admin?}
+  filter :user, as: :select, collection: proc {CompanyUsersCollectionQuery.new(current_user).call}
   filter :reviewed_at
   filter :created_at
 
   member_action :approve, method: :put, only: :index
   member_action :refuse, method: :put, only: :index
 
-  batch_action :refuse, if: proc { params[:scope] != "recusado" && params[:scope] != "aprovado" } do |ids|
```
adbatista, line 15: please try to keep unrelated changes out of the PR

app/admin/contributions.rb
```diff
@@ -51,7 +54,9 @@
     column :state do |contribution|
       Contribution.human_attribute_name("state/#{contribution.state}")
     end
-    column :reviewed_by
+    column :reviewed_by do |contribution|
```
adbatista, line 57: ` column :reviewed_by, &:reviewed_by_short_name`

app/decorators/contribution_decorator.rb
```diff
@@ -0,0 +1,20 @@
+class ContributionDecorator < Draper::Decorator
+  delegate_all
+
+  def user_short_name
+    user.name.split.values_at(0, -1).uniq.join(' ')
```
adbatista, line 5: this should be on the user model, or user decorator, is not concern of the contribution
so you could only use a delegate here

app/decorators/contribution_decorator.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def created_at
+    model.created_at.strftime("%d/%m/%Y")
+  end
+
+  def reviewed_by_short_name
+    reviewed_by.name.empty? ? '' : reviewed_by.name.split.values_at(0, -1).uniq.map(&:first).join.upcase
```
adbatista, line 13: same thing here

spec/decorators/contribution_decorator_spec.rb
```diff
@@ -0,0 +1,89 @@
+require 'rails_helper'
+
+RSpec.describe ContributionDecorator do
+
+  describe '#user_short_name' do
+    subject {described_class.new(contribution).user_short_name}
```
adbatista, line 6: don't add the exercise step in the subject, avoid this, subject should be only build of a object

adbatista (changes requested):

.gitignore
```diff
@@ -46,3 +46,5 @@ yarn-debug.log*
 .yarn-integrity
 public/webpack/*
 !public/webpack/.keep
+
+.idea/
```
adbatista, line 50: this should be in your global gitignore

spec/decorators/contribution_decorator_spec.rb
wenderjean: I disagree here :stuck_out_tongue_closed_eyes: 
It is pretty common to make the subject points to the method you're actually exercising on

app/decorators/contribution_decorator.rb
wenderjean: That's why I prefer to think of draper decorators as presenters that's what they actually are, I mean, this rule is specific to the contributions page, isn't it?

spec/decorators/contribution_decorator_spec.rb
adbatista: it's common but not a good practice

wenderjean: That's the reason we use the less descriptive version of subject for single-line exercises but for complex triggers, we usually give it a descriptive name. The point here is the format we usually expect for the file, I personally expect to see a `describe` block with a subject pointing to the method I'm dealing with.

```ruby
describe '#my_method' do
  subject { described_class.new.my_method }
  
  it { is_expected.to eq 10 }
end

describe 'GET /users' do
  subject(:make_request) { get '/users' }
  
  it 'expects something to happen' do
    make_request
    
    expect(response).to eq(...)
  end
end
```

But it's just my point, you can suggest any other format. :heart: :heart: :heart:

spec/decorators/contribution_decorator_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    context 'when user name is empty' do
+      let(:user) {build_stubbed(:user, name: '')}
+
+      it {is_expected.to eq ''}
+    end
+
+  end
+
+  describe '#reviewed_by_name' do
```
wenderjean, line 36: It should be `reviewed_by_short_name`

app/admin/contributions.rb
```diff
@@ -1,23 +1,24 @@
 ActiveAdmin.register Contribution do
+  decorate_with ContributionDecorator
   permit_params :state, :link, :user_id, :company_id, :repository_id
   actions :index, :show, :new, :create
 
   menu parent: Contribution.model_name.human(count: 2), priority: 1
 
-  filter :company, as: :select, if: proc { current_user.super_admin? }
-  filter :user, as: :select, collection: proc { CompanyUsersCollectionQuery.new(current_user).call }
+  filter :company, as: :select, if: proc {current_user.super_admin?}
```
wenderjean, line 8: Your editor is probably configured to remove spaces between `{}`, you should try to install editorconfig plugin on it.

spec/decorators/contribution_decorator_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  describe '#reviewed_at' do
+    subject {described_class.new(contribution).reviewed_at}
+
+    let(:contribution) {build_stubbed(:contribution, reviewed_at: DateTime.parse('2022-08-26T16:50'))}
+
+    it {is_expected.to eq '26/08/2022'}
+  end
+
```
wenderjean, line 88: Can we also remove these blank lines after describe blocks? :)

wenderjean (changes requested):

app/decorators/contribution_decorator.rb
wenderjean, line 13: It seems like `User#name` is always required to be present even though the `reviewed_by` itself is optional on the Contribution model. Maybe what you want is only `reviewed_by&.name.split.values_at(0, -1).uniq.map(&:first).join.upcase`?

app/decorators/contribution_decorator.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def reviewed_at
+    model.reviewed_at.strftime("%d/%m/%Y")
+  end
+
```
wenderjean, line 19: Can we remove this extra line?

app/decorators/contribution_decorator.rb
```diff
@@ -0,0 +1,19 @@
+class ContributionDecorator < Draper::Decorator
+  delegate_all
+
+  def created_at_formatted
```
wenderjean, line 4: I believe this suffix `_formatted` is kinda redundant considering we're dealing with a decorator, I mean, attributes coming from this object are supposed to be decorated already.

app/models/user.rb
```diff
@@ -170,4 +170,8 @@ def update_with_password(params, *options)
   def password_required?
     password_required
   end
+
+  def first_and_last_name
+    name.empty? ? '' : name.split.values_at(0, -1).uniq.join(' ')
```
wenderjean, line 175: You don't this ternary operator here, look
```ruby
''.split.values_at(0, -1).uniq.join(' ') => ''
```

spec/models/user_spec.rb
```diff
@@ -265,4 +265,26 @@
       expect(user).to_not have_admin_access
     end
   end
+
+  describe '#first_and_last_name' do
```
wenderjean, line 269: When @adbatista claimed about the use of the subject he was referring to something like this:
```ruby
describe '#first_and_last_name' do
  subject(:user) { build_stubbed(:user, name: name) } # Setup step
  
  context 'when name has 3 parts' do
    let(:name) { 'Name1 Name2 Name3' }
    
    it 'returns only first and last name' do
      expect(user.first_and_last_name).to eq 'Name1 Name3' # Verify step
    end
  end
end
```

app/decorators/contribution_decorator.rb
```diff
@@ -0,0 +1,19 @@
...
+  def reviewed_by_short_name
+    return '' if reviewed_by.nil?
+
+    reviewed_by.first_and_last_name.split.map(&:first).join.upcase
+  end
+
+  def reviewed_at_formatted
+    return '' if reviewed_at.nil?
+
+    model.reviewed_at.strftime("%d/%m/%Y")
```
wenderjean, line 17: You can also do something like `model.reviewed_at.to_fs(:date)`, it is actually a better approach once it'll be using the convention defined for the entire project, look [here](https://github.com/Codeminer42/Punchclock/blob/master/config/initializers/date_time_formats.rb#L3).

wenderjean (changes requested):

wenderjean (dismissed):

wenderjean (approved):

Issue #167, opened 2022-08-24
Change the permissions approach to be role-based


adbatista:
right now the user can have only one role, we need to change it to allow more than one role

PR #176, opened 2022-08-31
Adds multiple roles to users

mateuscelio:
https://github.com/Codeminer42/Punchclock/issues/167


adbatista:
why you did not use an enum for this?

app/models/user.rb
```diff
@@ -65,8 +68,13 @@ class User < ApplicationRecord
   scope :by_skills_in,   ->(*skill_ids) { UsersBySkillsQuery.where(ids: skill_ids) }
   scope :not_in_experience, -> { where arel_table[:created_at].lt(EXPERIENCE_PERIOD.ago) }
   scope :with_level,       -> value { where(level: value) }
+  scope :by_roles_in, lambda { |roles|
+    roles_values = self.roles.find_values(*roles).map(&:value)
```
GabrielRMuller, line 72: I think you can use `User.roles.values` in here

db/schema.rb
```diff
@@ -48,10 +48,10 @@
   end
 
   create_table "companies", id: :serial, force: :cascade do |t|
-    t.string "name", limit: 255
```
GabrielRMuller, line 51: just wondering, what is causing this change to the schema?

app/models/user.rb
mateuscelio: How exactly you suggest? `#find_values` are being used here to find only valid role name

app/admin/user.rb
```diff
@@ -196,11 +197,11 @@
         f.input :office
         f.input :company
         f.input :reviewer
-        f.input :role, as: :select, collection: User.role.values.map { |role| [role.text.titleize, role] }
+        f.input :roles, as: :check_boxes, collection: User.roles.values.map { |role| [role.titleize, role] }
```
adbatista, line 200: you can keep select, just change to accept multiple selections

app/admin/user.rb
```diff
@@ -196,11 +197,11 @@
...
         f.input :skills, as: :check_boxes
       else
         f.input :office, collection: current_user.company.offices.order(:city)
-        f.input :role, as: :select, collection: User.role.values.reject{ |value| value == 'super_admin' }.map { |role| [role.text.titleize, role] }
+        f.input :roles, as: :check_boxes, collection: Roles.all.reject { |role| role == :super_admin }.map { |role| [role.text.titleize, role] }
```
adbatista, line 204: same here

app/models/user.rb
```diff
@@ -55,6 +57,7 @@ class User < ApplicationRecord
   validates :email, uniqueness: true, presence: true
   validates :level, :specialty, presence: true, if: :engineer?
   validates :github, uniqueness: true, if: :engineer?
+  validates :roles, presence: true
```
adbatista, line 60: now we have multiple roles we don't need this to be required, if the user has no roles, this will be the normal

spec/models/user_spec.rb
```diff
@@ -100,6 +102,20 @@
                                               hr: 5) }
   end
 
+  describe 'roles' do
+    it { is_expected.to enumerize(:roles)
```
adbatista, line 106: use do..end sintax when multiple line block

spec/models/user_spec.rb
```diff
@@ -117,6 +133,39 @@
         expect(User.not_in_experience).to contain_exactly(user_not_in_experience)
       end
     end
+
+    describe '#by_roles_in' do
+      let!(:user1) { create(:user, roles: %i[admin super_admin]) }
+      let!(:user2) { create(:user, roles: [:super_admin]) }
+
+      subject { User.by_roles_in(roles_names).to_a }
```
adbatista, line 141: the subject should be only the build of the object being tested, not the result of the exercised method, this should be  in the test itself

spec/models/user_spec.rb
```diff
@@ -265,4 +314,28 @@
...
+
+    subject { user.is_admin? }
+    context 'with role hr' do
+      let(:role_name) { :hr }
+      it 'is considered an admin' do
+        is_expected.to eq(true)
+      end
+    end
+    context 'with role admin' do
+      let(:role_name) { :admin }
```
adbatista, line 329: do the build of the user here

spec/models/user_spec.rb
```diff
@@ -265,4 +314,28 @@
       expect(user).to_not have_admin_access
     end
   end
+
+  describe '#is_admin?' do
+    let(:user) { create(:user, roles: [role_name]) }
```
adbatista, line 319: this should be the subject

spec/models/user_spec.rb
```diff
@@ -265,4 +314,28 @@
...
+    context 'with role admin' do
+      let(:role_name) { :admin }
+      it 'is considered an admin' do
+        is_expected.to eq(true)
+      end
+    end
+    context 'with role super admin' do
+      let(:role_name) { :super_admin }
+      it 'is considered an admin' do
+        is_expected.to eq(true)
```
adbatista, line 337: `expect(subject.is_admin?).to be_truthy` or `expect(user.is_admin?).to be_truthy`

spec/models/user_spec.rb
```diff
@@ -117,6 +133,37 @@
         expect(User.not_in_experience).to contain_exactly(user_not_in_experience)
       end
     end
+
+    describe '#by_roles_in' do
```
adbatista, line 137: this is a scope wich is a class method, so the correct notation is `.by_roles_in` instead `#by_roles_in`

tonyaraujop: May I change all the other scope methods in this PR too?

adbatista: yes pls

PR #185, opened 2022-09-02
Add city reference to user

albarnaz:
Problem description: Sometimes the user didn't live in the same city of his office. So we need to indicate the city where the user lives.

This PR creates a structure to indicate the city where the user lives.

To create a new state just add it to the `create_sates.rake`

To create a new city just add it to the `create_cities.rake`


spec/factories/states.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :state do
+    name { "MyString" }
+    code { "MyString" }
+  end
```
adbatista, line 5: use sequence or faker here to create different states

spec/factories/users.rb
```diff
@@ -14,6 +14,7 @@
     company
     otp_required_for_login { false }
     office                { create(:office, company: company) }
+    city                  { create(:city) }
```
adbatista, line 17: you don't need to pass the block for relation

db/schema.rb
```diff
@@ -38,6 +38,14 @@
     t.index ["question_id"], name: "index_answers_on_question_id"
   end
 
+  create_table "cities", force: :cascade do |t|
+    t.string "name"
```
adbatista, line 42: can we have the same city name for the same state?

db/schema.rb
```diff
@@ -207,6 +215,13 @@
     t.index ["user_id"], name: "index_skills_users_on_user_id"
   end
 
+  create_table "states", force: :cascade do |t|
```
adbatista, line 218: state should be uniq

spec/factories/states.rb
albarnaz: done

spec/factories/users.rb
albarnaz: done

db/schema.rb
albarnaz: done

db/schema.rb
albarnaz: done

db/migrate/20220826124130_create_states.rb
```diff
@@ -0,0 +1,10 @@
+class CreateStates < ActiveRecord::Migration[6.1]
+  def change
+    create_table :states do |t|
```
adbatista, line 3: i forgot to add in the past review, Punch is Multitenant, so we should scope this to the company

lib/tasks/create_sates.rake
```diff
@@ -0,0 +1,42 @@
...
+      {name: 'Roraima', code: 'RR'},
+      {name: 'Santa Catarina', code: 'SC'},
+      {name: 'São Paulo', code: 'SP'},
+      {name: 'Sergipe', code: 'SE'},
+      {name: 'Tocantins', code: 'TO'}
+    ]
+
+    ActiveRecord::Base.transaction do
+      states.each do |st|
+        State.find_or_create_by!(code: st[:code]) do |state|
```
adbatista, line 36: we need to pass the company as argument

carloslopes:
Guys, beyond this feature, we must remove the Offices part from the project. We aren't based in offices anymore.

Here are the things that we should implement together with what was already done here:
- [x] Add city to users
- [ ] Remove offices from the project
- [ ] Add `Head` field on the users, where we should be able to point who is the head of that employee

Let me know what you think about this

albarnaz:
Guys, sorry for the delay. I'm changing my client, so I was in the last week and focused on delivering the last issues. I will be back on this issue soon

albarnaz:
@carloslopes I think it makes sense, but we can break it into issues, so, in my opinion, we can first finish the city of the user, then we can implement the head of the user, and then we can remove the office.
What do you think about it?

carloslopes:
Agree!

PR #188, opened 2022-09-02
Allocation bug

### LINKED ISSUE #187: Allocation chart does not show active allocation
When user has more than one allocation, one active and the other in the future. The allocation chart page shows only the inactive allocaction, that will happen in the future.

tonyaraujop:
Closes #187 

Allocation page:
image

Allocation chart:
image

### DIFF
--- a/app/queries/allocations_and_unalocated_users_query.rb
+++ b/app/queries/allocations_and_unalocated_users_query.rb
@@ -30,6 +30,6 @@ class AllocationsAndUnalocatedUsersQuery
   end
 
   def last_user_allocations
-    Allocation.group(:user_id).maximum(:end_at).values
+    Allocation.where(ongoing: true).group(:user_id).maximum(:end_at).values
   end
 end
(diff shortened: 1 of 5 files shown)


app/queries/allocations_and_unalocated_users_query.rb
```diff
@@ -30,6 +30,6 @@ def where
   end
 
   def last_user_allocations
-    Allocation.group(:user_id).maximum(:end_at).values
+    Allocation.where(ongoing: true).group(:user_id).maximum(:end_at).values
```
adbatista, line 33: why not use the allocation scope, `.ongoing` ?

spec/models/allocation_spec.rb
```diff
@@ -176,5 +176,13 @@
         expect(subject.days_until_finish).to eq('Finalizado')
       end
     end
+
+    context 'when allocation did not started' do
+      subject { build_stubbed(:allocation, end_at: Time.zone.today + 1.month) }
```
adbatista, line 181: just to let you know; you can also do `1.month.from_now`

app/queries/allocations_and_unalocated_users_query.rb
tonyaraujop: I miss that one. Changing right now.

spec/models/allocation_spec.rb
tonyaraujop: Nice.

PR #189, opened 2022-09-02
Adds link to allocation in column "Alocado até" in allocation chart

mateuscelio:
https://github.com/Codeminer42/Punchclock/issues/179

image

### DIFF
--- a/app/admin/allocation_chart.rb
+++ b/app/admin/allocation_chart.rb
@@ -39,7 +39,8 @@ def build_allocation_status_cell(last_allocation)
       last_allocation: last_allocation,
       allocation_status: allocation_status
     ),
-    allocation_status: allocation_status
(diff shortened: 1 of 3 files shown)


app/admin/allocation_chart.rb
```diff
@@ -39,7 +39,8 @@ def build_allocation_status_cell(last_allocation)
       last_allocation: last_allocation,
       allocation_status: allocation_status
     ),
-    allocation_status: allocation_status
+    allocation_status: allocation_status,
+    last_allocation: last_allocation
```
adbatista, line 43: Why?

mateuscelio: Because I use this value to generate the path in `admin_allocation_path`

GabrielRMuller (approved):

PR #190, opened 2022-09-05
Feature/present only first and last name on allocation chart

### LINKED ISSUE #173: Present only first and last name on allocation chart
We should present only user's first and last name on the allocation chart page.

vinas-go:
Resolve #173

### DIFF
--- a/app/admin/contributions.rb
+++ b/app/admin/contributions.rb
@@ -88,7 +88,7 @@ ActiveAdmin.register Contribution do
       if current_user.super_admin?
         f.input :company
       else
-        f.input :company_id, as: :hidden, input_html: { value: current_user.company_id }
+        f.input :company_id, as: :hidden, input_html: {value: current_user.company_id}
       end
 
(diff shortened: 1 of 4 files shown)


app/admin/contributions.rb
```diff
@@ -88,7 +88,7 @@
       if current_user.super_admin?
         f.input :company
       else
-        f.input :company_id, as: :hidden, input_html: { value: current_user.company_id }
+        f.input :company_id, as: :hidden, input_html: {value: current_user.company_id}
```
wenderjean, line 91: Can we revert this change? :nail_care:

spec/models/user_spec.rb
```diff
@@ -371,5 +371,41 @@
         end
       end
     end
+
+    describe '#first_and_last_name' do
```
wenderjean, line 375: This change is on master already, I don't think it should be present in this PR :thinking:

adbatista:
@vinas-go what is the status of this

adbatista:
closing this PR seems to be abandoned, I'll address it to someone else

PR #192, opened 2022-09-05
Allocations page improvements

carloslopes:


GabrielRMuller:
We got a failing spec

app/admin/allocations.rb
```diff
@@ -1,36 +1,25 @@
...
-  scope :all do
-    current_user.company.allocations.all.joins(:user).merge(User.active)
-  end
-  scope :spreadsheet do |relation|
-    AllocationsAndUnalocatedUsersQuery.new(relation, current_user.company).call
-  end
-
+  filter :ongoing
   filter :user, collection: proc {
-    current_user.super_admin? ? User.all.order(:name).group_by(&:company) : current_user.company.users.order(:name)
```
GabrielRMuller, line 21: just curious, why do we want to remove this `group_by`?

carloslopes: Because it causes a bug when the user accessing the page is a super admin. If the dropdown has subgroups of companies, it generates options like this one: `<option value="name">name</option>`.

GabrielRMuller (approved):

PR #194, opened 2022-09-05
Add market info on projects

carloslopes:


carloslopes:
The issues pointed by CodeClimate are false positives

db/migrate/20220905200117_add_market_to_projects.rb
```diff
@@ -0,0 +1,5 @@
+class AddMarketToProjects < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :market, :integer
```
tonyaraujop, line 3: Why not to also add for this migration `null: false` in order to have validation in database level?

app/models/project.rb
```diff
@@ -1,17 +1,21 @@
 # frozen_string_literal: true
 
 class Project < ApplicationRecord
+  enum :market, %i[internal international]
```
adbatista, line 4: we have enumerize in the project that already supports i18n, please use it to keep consistent with other models of the project

adbatista (dismissed):

adbatista:
can you solve the conflicts please?

PR #199, opened 2022-09-08
Add date order filter

### LINKED ISSUE #200: Order date filter
Fix order date filter on allocations page

tonyaraujop:
Closes #200

Date order filter
[Gravação de tela de 08-09-2022 14:53:49.webm](https://user-images.githubusercontent.com/92229784/189196210-402be4e7-19b3-44d8-81d7-192017ecc706.webm)

### DIFF
--- a/app/admin/allocation_chart.rb
+++ b/app/admin/allocation_chart.rb
@@ -22,6 +22,9 @@ ActiveAdmin.register_page 'Allocation Chart' do
         column(I18n.t('skills'), class: 'allocated-column__last') do |allocation|
           decorated_user(allocation).skills
         end
+        column(:allocated_until_data) do |allocation|
+          allocation.end_at ? allocation.end_at.to_time.to_i : 'Not allocated'
+        end
       end
     end
   end
--- a/app/assets/javascripts/active_admin.js
+++ b/app/assets/javascripts/active_admin.js
@@ -4,6 +4,7 @@
 //= require stats.js
 //= require chartkick
 //= require Chart.bundle
+//= require moment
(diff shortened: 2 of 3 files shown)


app/admin/allocation_chart.rb
```diff
@@ -22,6 +22,9 @@
         column(I18n.t('skills'), class: 'allocated-column__last') do |allocation|
           decorated_user(allocation).skills
         end
+        column(:allocated_until_data) do |allocation|
+          allocation.end_at ? allocation.end_at&.to_time.to_i : 'Not allocated'
```
GabrielRMuller, line 26: do we need the safe operator if we're checking if it exists first?

package.json
```diff
@@ -28,7 +28,7 @@
     "immutable": "~3.8.1",
     "jquery": "^3.2.1",
     "lodash": "~4.17.4",
-    "moment": "~2.29.4",
+    "moment": "^2.29.4",
```
GabrielRMuller, line 31: just curious, why did this change?

tonyaraujop: I think this happens because I reinstalled this lib. But We can leave it the way it was.

app/admin/allocation_chart.rb
tonyaraujop: Actually no. I'll remove right now.

package.json
GabrielRMuller: if this is unrelated to this work I think we should leave the way it was.

GabrielRMuller (approved):

PR #201, opened 2022-09-09
Upgrade ruby to 3.1

adbatista:


wenderjean (dismissed):
LGTM

Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN mkdir -p ${HOME} && \
 USER devel:devel
 WORKDIR $APP
 
-RUN gem install bundler -v 2.1.4
+RUN gem install bundler
```
wenderjean, line 52: We don't need to change this but there is an option I like to use that is `gem install bundler -v $(tail -n 1 Gemfile.lock)`

wenderjean (approved):

PR #205, opened 2022-09-12
Revenue forecast

carloslopes:
This PR implements the Revenue Forecast info, where we can view the revenue forecast of a specific allocation, project, or of everything on the Revenue Forecast panel page.

Screen Shot 2022-09-12 at 16 46 20
Screen Shot 2022-09-12 at 16 46 36
Screen Shot 2022-09-12 at 16 55 12


app/admin/project.rb
```diff
@@ -76,6 +76,30 @@
             end
           end
         end
+
+        panel t('revenue_forecast') do
+          data = RevenueForecastService.project_forecast(project)
+
+          # TODO: Refactor
```
GabrielRMuller, line 83: would be nice if we already added this TODO's as issues on the repo.

app/services/revenue_forecast_service.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  def calculate_forecast(working_days, hourly_rate, beginning_of_month)
+    exchange_rate_date = if beginning_of_month > Date.current
+                           Date.current.beginning_of_month - 1
+                         else
+                           beginning_of_month - 1
+                         end
+
+    converted_rate = hourly_rate.exchange_to_historical('BRL', exchange_rate_date)
+    converted_rate * working_days * 8
```
GabrielRMuller, line 102: is `8` the working hours? I'd suggest adding this value to a constant or a variable so it describes what it is

PR #207, opened 2022-09-13
Allocations can't be endless anymore

carloslopes:


GabrielRMuller (approved):
LGTM! 

I'm wondering if there is any changes necessary to the seed file

PR #208, opened 2022-09-13
Mentoring screen

### LINKED ISSUE #203: Create a new Screen to list mentors and mentees in the admin
We would like to have three ways to view the list of mentors and mentees:

- a table of everybody that has a mentor: `mentor | mentee`
- a list of mentors grouped by their offices (each mentor with their respective mentee)
- a list of mentees grouped by their offices (each mentee with their respective mentor)

Those lists by office could be either on the office page or a separate page. Being in a separate page would be good to see all of the groups at once instead of having to access the page of each office.

If possible, it would also be good to see who are the mentees of a given user when we access the user profile (the same way we're already able to see who is the mentor of a user when we access their profile today)

tonyaraujop:
closes #203 

Mentoring screen:
image

### DIFF
--- a/app/admin/mentoring.rb
+++ b/app/admin/mentoring.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+ActiveAdmin.register_page 'Mentoring' do
+  menu label: proc { I18n.t('mentoring') }, parent: User.model_name.human(count: 2)
+
+  content title: I18n.t('mentoring') do
+    tabs do
+      mentoring = MentoringQuery.new
+      tab I18n.t('mentoring') do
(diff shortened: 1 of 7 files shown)


app/admin/mentoring.rb
```diff
@@ -10,7 +10,7 @@
         mentors = mentoring.call
         table_for mentors do
           column(I18n.t('mentor')) { |mentor| mentor.mentors }
-          column(I18n.t('office')) { |office| office.city }
+          column(:office) { |office| office.city }
```
adbatista, line 13: ❤️

GabrielRMuller (approved):

PR #209, opened 2022-09-14
Add offices tab to mentoring page

tonyaraujop:
This PR implements a second tab to `/admin/mentoring` page.

image


app/queries/mentoring_query.rb
```diff
@@ -13,7 +13,8 @@ def call
       .select(
         'array_agg(users.name) as mentees',
         'reviewers_users.name as mentors',
-        :city
+        :city,
+        "count('x') OVER (PARTITION BY offices.city) as city_count"
```
GabrielRMuller, line 17: what is this part of the query doing exactly?

app/views/admin/mentoring/_offices_mentoring.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      <th><%= I18n.t('offices') %></th>
+      <th><%= I18n.t('mentors') %></th>
+      <th><%= I18n.t('mentees') %></th>
+    </tr>
+  </thead>
+  <tbody>
+    <% city_name = nil %>
+    <% mentorings.each do |mentoring| %>
+      <% city = mentoring.city == city_name ? nil : mentoring.city %>
+      <% city_name = mentoring.city %>
```
GabrielRMuller, line 13: sorry not sure if I understood correctly, are you trying to group the mentors by city in here?

app/queries/mentoring_query.rb
tonyaraujop: Only to count how many times a city appears, so it can be used to span table row.

app/views/admin/mentoring/_offices_mentoring.html.erb
tonyaraujop: Basically yes and also to remove the repeated city names.

app/views/admin/mentoring/_offices_mentoring.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <tbody>
+    <% mentorings.each do |city, mentoring| %>
+      <tr>
+        <th class="mentoring_table__mentor_city" rowspan="<%= mentorings.count %>"><%= city %></th>
+      <% mentoring.each do |mentor| %>
+        <td class="mentoring_table__mentor_name"><%= mentor.name %></td>
+        <td class="mentoring_table__mentor_mentees"><%= mentor.mentees.join(', ') %></td>
+      </tr>
+      <% end %>
+    <% end %>
```
GabrielRMuller, line 18:
```suggestion
    <% mentorings.each do |city, mentoring| %>
      <tr>
        <th class="mentoring_table__mentor_city" rowspan="<%= mentorings.count %>"><%= city %></th>
        <% mentoring.each do |mentor| %>
          <td class="mentoring_table__mentor_name"><%= mentor.name %></td>
          <td class="mentoring_table__mentor_mentees"><%= mentor.mentees.join(', ') %></td>
        <% end %>
      </tr>
    <% end %>
```

GabrielRMuller (commented):
Can we add some different cases in MentoryQuerySpec? maybe one with multiple mentees and one without mentors, etc.

tonyaraujop:
> Can we add some different cases in MentoryQuerySpec? maybe one with multiple mentees and one without mentors, etc.

Sure. This can be done.

spec/queries/mentoring_query_spec.rb
```diff
@@ -6,25 +6,74 @@
...
 
-    context 'when there are mentees and mentors' do
+    context 'when there are one mentee and mentor' do
       let(:mentor) { create(:user) }
       let!(:mentee) { create(:user, reviewer_id: mentor.id) }
 
-      let(:mentors) { call.map(&:mentors) }
+      let(:mentors) { call.map(&:name) }
       let(:offices) { call.map(&:city) }
       let(:mentees) { call.map(&:mentees) }
```
GabrielRMuller, line 15: although DRY is not always the way to go when it comes to testing, you could have these 3 calls outside of the `contexts` and leave it under the `describe` scope, so all the contexts can have access to it.

spec/queries/mentoring_query_spec.rb
```diff
@@ -6,25 +6,74 @@
...
+      it "returns mentor's mentees" do
+        expect(mentees.flatten).to contain_exactly(bob_mentee.name, other_bob_mentee.name, brown_mentee.name, other_brown_mentee.name)
+      end
+    end
+    
+    context 'when there are no mentees or mentor' do
+      let!(:bob) { create(:user) }
+      let!(:brow) { create(:user) }
+      let!(:john) { create(:user) }
+      let(:joe) { create(:user) }
```
GabrielRMuller, line 60: since we're not going to use any of these values separately, we can create them and not assign to any variable:
```suggestion
      before do
        4.times { create(:user) }
      end
```

GabrielRMuller (changes requested):
Also noticed there were some trailing whitespaces on the spec file, so watch out for that :eyes:

spec/queries/mentoring_query_spec.rb
tonyaraujop: Nice!

spec/queries/mentoring_query_spec.rb
tonyaraujop: > Also noticed there were some trailing whitespaces on the spec file, so watch out for that eyes

I'll pass rubocop one more time in those files.

GabrielRMuller (dismissed):

GabrielRMuller (approved):

PR #212, opened 2022-09-16
Fix mentoring screen

tonyaraujop:
Office mentoring screen:
image

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -323,3 +323,12 @@ contributions = create_contributions(
   repositories: repos,
   dates: contributions_dates
 )
+
+print 'Creating mentors to some users..'
+
+30.times do
+  mentor = User.all.sample
+  mentee = User.all.sample.update(reviewer_id: mentor.id)
+end
+
+puts 'Done.'
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -328,7 +328,7 @@ print 'Creating mentors to some users..'
 
 30.times do
   mentor = User.all.sample
-  mentee = User.all.sample.update(reviewer_id: mentor.id)
+  User.all.sample.update(reviewer_id: mentor.id)
 end
 
 puts 'Done.'
--- a/spec/queries/mentoring_query_spec.rb
+++ b/spec/queries/mentoring_query_spec.rb
@@ -51,21 +51,11 @@ RSpec.describe MentoringQuery do
       end
 
       context 'and mentee are inactive' do
-        before do
-          bob_mentee.update(active: false)
-          brown_mentee.update(active: false)
-        end
+        let!(:inactive_bob_mentee) { create(:user, reviewer_id: brown.id, active: false) }
+        let!(:inactive_brown_mentee) { create(:user, reviewer_id: brown.id, active: false) }
 
-        it 'returns mentors' do
-          expect(mentors).to contain_exactly(bob.name, brown.name)
-        end
-
-        it "returns mentors's offices" do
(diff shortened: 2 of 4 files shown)


db/seeds.rb
```diff
@@ -323,3 +323,12 @@ def create_repository(link:, company:)
   repositories: repos,
   dates: contributions_dates
 )
+
+print 'Creating mentors to some users..'
+
+30.times do
+  mentor = User.all.sample
+  mentee = User.all.sample.update(reviewer_id: mentor.id)
```
adbatista, line 331: What is the need for this assign?

spec/queries/mentoring_query_spec.rb
```diff
@@ -36,16 +36,37 @@
       let!(:brown_mentee) { create(:user, reviewer_id: brown.id) }
       let!(:other_brown_mentee) { create(:user, reviewer_id: brown.id) }
```
adbatista, line 37: does these lets need to be eager?

spec/queries/mentoring_query_spec.rb
```diff
@@ -36,16 +36,37 @@
...
+          expect(mentees.flatten).to contain_exactly(bob_mentee.name, other_bob_mentee.name, brown_mentee.name, other_brown_mentee.name)
+        end
       end
 
-      it "returns mentors's offices" do
-        expect(offices).to contain_exactly(bob.office.city, brown.office.city)
-      end
+      context 'and mentee are inactive' do
+        before do
+          bob_mentee.update(active: false)
```
adbatista, line 55: is better to create an inactive, with this update you are hitting de DB twice

db/seeds.rb
tonyaraujop: You're right. There is not need for this `mentee` assign.

spec/queries/mentoring_query_spec.rb
tonyaraujop: Yes, because when the method is called, it search for mentees in database, if they are not created with `!`, the test will fail.

spec/queries/mentoring_query_spec.rb
tonyaraujop: I'll do this changes.

spec/queries/mentoring_query_spec.rb
adbatista: so put this inside the context that is using it

spec/queries/mentoring_query_spec.rb
```diff
@@ -29,23 +29,34 @@
 
     context 'when there are multiple mentees and mentors' do
       let(:bob) { create(:user) }
-      let!(:bob_mentee) { create(:user, reviewer_id: bob.id) }
-      let!(:other_bob_mentee) { create(:user, reviewer_id: bob.id) }
-
       let(:brown) { create(:user) }
```
GabrielRMuller, line 32: I'd suggest we switch these names to something more self-explanatory (`:mentor_1`, `:mentor_2`). It will make easier to understand what we want from the expectations.

db/seeds.rb
```diff
@@ -323,3 +323,12 @@ def create_repository(link:, company:)
   repositories: repos,
   dates: contributions_dates
 )
+
+print 'Creating mentors to some users..'
+
+30.times do
+  mentor = User.all.sample
+  User.all.sample.update(reviewer_id: mentor.id)
```
GabrielRMuller, line 331: I know that performance on the seed file is not that big of a deal, but I suggest we pass this `User.all` to a variable before using it on lines 330 and 331 so we don't end up pulling all the Users two times.

Issue #233, opened 2022-10-03
Update the roles permissions


fabiosammy:
Following the roles, update the permissions accordingly:

* normal: 0 - Should have the basic access to the punch and open source lists (is not a special access, remove it)
* evaluator: 1 - Should have access to the evaluation of the users
* admin: 2 - Should have all access, except the destroy cases
* open_source_manager: 3, - Remove
* hr: 4 - Should have finance access - Remove

PR #234, opened 2022-10-03
Update the roles permissions

fabiosammy:
Refers to #233

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -275,14 +275,4 @@ RSpec.describe User, type: :model do
     it { expect(inactive_user.inactive_message).to eq :inactive_account }
     it { expect(active_user.inactive_message).to eq :unconfirmed }
   end
-
-  describe '#has_admin_access' do
-    it "allows admin access for user with admin role" do
-      expect(admin_user).to have_admin_access
-    end
-
-    it "not allows admin access for user without admin roles" do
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -146,11 +147,11 @@ def office_head?
...
 
   def has_admin_access?
-    admin? || open_source_manager? || hr?
+    admin?
   end
 
   def is_admin?
-    admin? || hr?
+    admin?
   end
```
GabrielRMuller, line 155: can we change this to a single method now? or change it to an alias if you prefer

spec/features/admin/punches_spec.rb
```diff
@@ -52,7 +52,7 @@
   end
 end
 
-describe "Punches with normal admin_user", type: :feature do
+describe "Punches with a admin_user", type: :feature do
```
GabrielRMuller, line 55:
```suggestion
describe "Punches with an admin_user", type: :feature do
```

app/models/user.rb
fabiosammy: I was thinking about it, but not sure if we gonna need this with another kind of role, maybe for the evaluator kind? @adbatista ?

Also, seems that this is to check if the user can access the active admin, and some elements in the dashboard, but also it does not makes sense. Maybe in another PR.

GabrielRMuller (approved):

PR #267, opened 2022-10-11
Add Vacation model

### LINKED ISSUE #266: Add vacation Request
Allow users to request vacations

it should have the period when creating

when created it should have the status pending

the available statuses are pending, approved, and denied

tonyaraujop:
Closes #266

### DIFF
--- a/db/migrate/20221011125254_create_vacations.rb
+++ b/db/migrate/20221011125254_create_vacations.rb
@@ -1,0 +1,16 @@
+class CreateVacations < ActiveRecord::Migration[7.0]
+  PENDING_STATUS = 0
+
+  def change
+    create_table :vacations do |t|
+      t.date :start_date, null: false
+      t.date :end_date, null: false
+      t.integer :status, default: PENDING_STATUS
+      t.references :user, null: false, foreign_key: true
+      t.references :commercial_approver, foreign_key: { to_table: :users }
+      t.references :administrative_approver, foreign_key: { to_table: :users }
+
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_10_06_131621) do
+ActiveRecord::Schema[7.0].define(version: 2022_10_11_125254) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
@@ -48,7 +48,6 @@ ActiveRecord::Schema[7.0].define(version: 2022_10_06_131621) do
 
   create_table "contributions", force: :cascade do |t|
     t.bigint "user_id"
-    t.bigint "company_id"
     t.string "link", null: false
     t.string "state", null: false
     t.datetime "created_at", precision: nil, null: false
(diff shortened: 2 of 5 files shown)


db/migrate/20221011125254_create_vacations.rb
```diff
@@ -0,0 +1,14 @@
+class CreateVacations < ActiveRecord::Migration[7.0]
+  def change
+    create_table :vacations do |t|
+      t.date :start_date, null: false
+      t.date :end_date, null: false
+      t.integer :status, default: 1
```
rhian-cs, line 6: Per #266, shouldn't status start as "pending"?

Also, you could put this in a separate variable, e.g. `PENDING_STATUS = 0` and use `t.integer :status, default: PENDING_STATUS`

db/schema.rb
```diff
@@ -57,7 +56,6 @@
     t.bigint "reviewer_id"
     t.datetime "reviewed_at", precision: nil
     t.string "pr_state"
-    t.index ["company_id"], name: "index_contributions_on_company_id"
     t.index ["link"], name: "index_contributions_on_link", unique: true
```
rhian-cs, line 59: Should these other columns in unrelated tables be deleted from the schema?

Perhaps you need to run your migration on top of the existing schema (the one on `master`), but this is more of a guess.

spec/factories/vacations.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :vacation do
+    start_date { Date.today }
+    end_date { 2.months.from_now }
+    status { 1 }
```
rhian-cs, line 5:
```suggestion
    status { Vacation.statuses[:pending] }
```
I'm not sure if this will work but it would be best if you could just reference the status as `pending` without depending on the number.

I got the code from here: https://www.rubyinrails.com/2019/04/12/how-to-get-integer-value-from-enum-in-rails/

spec/factories/vacations.rb
```diff
@@ -0,0 +1,10 @@
...
+    user { nil }
```
rhian-cs, line 6:
```suggestion
    user
```

Since the `user` is required for a Vacation to be valid, you can just associate it in the factory, so that you won't have to do that for every `Vacation` object generated from the factory

db/migrate/20221011125254_create_vacations.rb
```diff
@@ -0,0 +1,18 @@
+class CreateVacations < ActiveRecord::Migration[7.0]
+  PENDING_STATUS = 0
+  APPROVED_STATUS = 1
+  DENIED_STATUS = 2
```
adbatista, line 4: why this if you are using only one?

tonyaraujop: This was an mistake. I'll remove now.

db/migrate/20221011125254_create_vacations.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def change
+    create_table :vacations do |t|
+      t.date :start_date, null: false
+      t.date :end_date, null: false
+      t.integer :status, default: PENDING_STATUS
+      t.references :user, null: false, foreign_key: true
+      t.references :commercial_approver, foreign_key: { to_table: :users }
+      t.references :admin_approver, foreign_key: { to_table: :users }
```
adbatista, line 13: administrative_approver

PR #269, opened 2022-10-13
Reimplement Allocations spreadsheet with caxlsx gem

FabioSaito:
Implements #248


spec/spreadsheets/allocations_spreadsheet_spec.rb
```diff
@@ -30,35 +30,6 @@
     ]
   end
 
-  describe '#to_string_io' do
```
adbatista, line 33: why are you removing it?

Issue #275, opened 2022-10-19
Create pages for vacations in the punch normal user side


fabiosammy:
For a "normal" user, create pages to see, request or cancel a vacation.

The user should be able to see all your own vacations requests
The normal user cannot cancel a not pending vacation

The form will have only the start date and end date.
The end date cannot be before the start date.
The start and the end date cannot be in the past.
The end date cannot be on the same day as the start date

HeitorMC:
Hello There!

I would like to be assigned to this issue

PR #282, opened 2022-10-24
Feature: Create pages for vacations in the punch normal user side

HeitorMC:
Closes: #275 

### What?

For a "normal" user, create pages to see, request or cancel a vacation.

The user should be able to see all your own vacations requests
The normal user cannot cancel a not pending vacation

The form will have only the start date and end date.
The end date cannot be before the start date.
The start and the end date cannot be in the past.
The end date cannot be on the same day as the start date

### DIFF
--- a/app/controllers/vacations_controller.rb
+++ b/app/controllers/vacations_controller.rb
@@ -1,0 +1,64 @@
+class VacationsController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
+    @vacations = scopped_vacations
+  end
+
+  def show
+    @vacation = scopped_vacations.find(params[:id])
+  end
+
+  def new
+    @vacation = Vacation.new
+  end
+
+  def create
+    @vacation = scopped_vacations.new(vacation_params)
+
+    if @vacation.save
+      redirect_to vacations_path, notice: I18n.t(:notice, scope: "flash.actions.create", resource_name: "Vacation")
+    else
+      flash_errors('create')
+      render :new
+    end
+  end
+
+  def cancel
+    vacation = Vacation.find(params[:id])
+
+    if vacation.pending?
+      vacation.update(status: :cancelled)
+
+      redirect_to vacations_path, notice: I18n.t(:notice, scope: "flash.vacation.update")
+    else
+      redirect_to vacations_path, alert: I18n.t(:alert, scope: "flash.vacation.update")
+    end
+  end
+
+  private
+
+  def scopped_vacations
+    current_user.vacations
+  end
+
+  def vacation_params
+    params.require(:vacation).permit(:start_date, :end_date)
+  end
+
+  def flash_errors(scope)
+    flash.now[:alert] = "#{alert_message(scope)} #{error_message}"
+  end
+
+  def alert_message(scope)
+    I18n.t(:alert, scope: "flash.actions.#{scope}", resource_name: "Vacation")
+  end
+
+  def errors
+    @vacation.errors.full_messages.join('. ')
+  end
+
+  def error_message
+    I18n.t(:errors, scope: "flash", errors: errors)
+  end
+end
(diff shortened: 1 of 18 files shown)


config/locales/en/en.yml
```diff
@@ -69,3 +69,6 @@ en:
   english_evaluation: "English evaluation"
   revenue_forecast: 'Revenue Forecast'
   revenue_forecast_warning: '* This feature was implemented on Sept/2022, and for this reason, data before this date are incomplete.'
+  vacations:
+    index:
+      are_you_sure_remove_request: 'Remove this vacation request?'
```
GabrielRMuller, line 74: [nit] need a new line

config/locales/miners/miner.en.yml
```diff
@@ -66,3 +66,9 @@ en:
         open: Open
         closed: Closed
         merged: Merged
+    vacation:
+      status:
+        pending: Pending
+        approved: Approved
+        refused: Refused
+        canceled: Canceled
```
GabrielRMuller, line 74: in this translation file we're calling it `canceled` with a single L, everywhere else we're calling `cancelled` with two L's, maybe change it to `cancelled` here as well?

spec/controllers/vacations_controller_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  before do
+    allow(controller).to receive(:authenticate_user!)
+    allow(controller).to receive(:current_user).and_return(user)
+  end
+
+  describe 'GET index' do
+    let!(:vacation) { create(:vacation, user: user) }
+
+    it "returns all user's vacations" do
```
GabrielRMuller, line 14: this spec is only checking if the request is successful, but the `it` suggests it's testing if all the user's vacations is being returned. I'd suggest renaming this one to reflect what it's testing and then add a new one to check if we're getting the vacations.

spec/controllers/vacations_controller_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      get :index
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe 'GET show' do
+    let(:vacation) { create(:vacation, user: user) }
+
+    it "returns a specific user vacation" do
```
GabrielRMuller, line 24: similar to my comment on index

app/models/vacation.rb
```diff
@@ -10,9 +11,33 @@ class Vacation < ApplicationRecord
...
+    @starting_day = date.presence
+    mount_start_date
+    date
+  end
+
+  def ending_day=(date)
+    @ending_day = date.presence
+    mount_end_date
+    date
+  end
```
GabrielRMuller, line 32: sorry I didn't quite understood what are these methods doing and where they're being used, could you elaborate a little bit on them?

app/controllers/vacations_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      vacation_to_cancel.update!(status: :cancelled)
+      redirect_to vacations_path, notice: I18n.t(:notice, scope: "flash.vacation.cancel")
+    else
+      redirect_to vacations_path, alert: I18n.t(:alert, scope: "flash.vacation.cancel")
+    end
+  end
+
+  private
+
+  def scopped_vacations
```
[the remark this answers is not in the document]
FabioSaito: Fixed! Thanks

app/views/vacations/index.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= link_to 'Solicitar Férias', new_vacation_path %>
```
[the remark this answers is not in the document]
FabioSaito: good catch, i misssed this one

spec/controllers/vacations_controller_spec.rb
FabioSaito: Could you review again? I just updated these tests

app/models/vacation.rb
FabioSaito: I reimplemented in a better way now, it was a mistake

GabrielRMuller (approved):
nice specs!

PR #313, opened 2022-11-29
Revenue forecast improvements

carloslopes:
This PR's objective is to change how the allocations' hourly rates are converted from US dollars to Brazilian Reais. Before this change, we were using a gem to provide the exchange rate, but we can't guarantee that the value provided by this source is reliable. Because of that, we decided to create our own database of rates and use it to do the calculations

### DIFF
--- a/db/migrate/20221125185846_create_exchange_rates.rb
+++ b/db/migrate/20221125185846_create_exchange_rates.rb
@@ -1,0 +1,13 @@
+class CreateExchangeRates < ActiveRecord::Migration[7.0]
+  def change
+    create_table :exchange_rates do |t|
+      t.integer :month, null: false
+      t.integer :year, null: false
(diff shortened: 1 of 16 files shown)


db/migrate/20221125185846_create_exchange_rates.rb
```diff
@@ -0,0 +1,13 @@
+class CreateExchangeRates < ActiveRecord::Migration[7.0]
+  def change
+    create_table :exchange_rates do |t|
+      t.integer :month, null: false
+      t.integer :year, null: false
+      t.string :currency, null: false
+      t.float :rate, null: false
```
adbatista, line 7: maybe is better to use decimal here to avoid float precision issue

fabiosammy (approved):

fabiosammy (approved):

PR #319, opened 2022-12-07
Fix: Minimum days for vacation

HeitorMC:
As mentioned in #318, this PR contemplates a new error message for vacations that are less than 10 days.

image

### DIFF
--- a/spec/models/vacation_spec.rb
+++ b/spec/models/vacation_spec.rb
@@ -72,6 +72,28 @@ RSpec.describe Vacation, type: :model do
 
       it { expect(vacation).to_not be_valid }
     end
+
+    context 'when the duration of the vacation is less than 10 days' do
+      let(:vacation) {build(:vacation, start_date: 1.day.from_now, end_date: 2.days.from_now)}
+
+      it { expect(vacation).to_not be_valid }
+
+      it 'raises an error' do
+        expect { vacation.save! }.to raise_error(ActiveRecord::RecordInvalid, 'A validação falhou: As férias precisam ter no mínimo 10 dias')
+      end
+    end
+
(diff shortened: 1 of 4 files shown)


spec/models/vacation_spec.rb
```diff
@@ -72,6 +72,16 @@
 
       it { expect(vacation).to_not be_valid }
     end
+
+    context 'when the duration of the vacation is less than 10 days' do
```
adbatista, line 76: and when vacation duration is equal 10 days and more than 10 days?

spec/models/vacation_spec.rb
```diff
@@ -72,6 +72,28 @@
...
+
+      it { expect(vacation).to_not be_valid }
+
+      it 'raises an error' do
+        expect { vacation.save! }.to raise_error(ActiveRecord::RecordInvalid, 'A validação falhou: As férias precisam ter no mínimo 10 dias')
+      end
+    end
+
+    context 'when the duration of the vacation is equal or higher than 10 days' do
+      let(:vacation) {build(:vacation, start_date: 1.day.from_now, end_date: 11.days.from_now)}
```
adbatista, line 87: what about exactly 10 days?

spec/models/vacation_spec.rb
HeitorMC: Done

spec/models/vacation_spec.rb
HeitorMC: Alright, now we have specs to cover all the three cases: lower than 10 days, equal to 10 days, and higher than 10 days covered

adbatista (approved):

PR #329, opened 2022-12-19
Revenue forecast adjustments

carloslopes:


app/presenters/revenue_forecast_presenter.rb
```diff
@@ -6,8 +6,8 @@ class RevenueForecastPresenter
   # should only show forecasts from 2022 and beyond
   REVENUE_FORECAST_START_YEAR = 2022
 
-  def initialize(year)
-    @forecasts = RevenueForecastService.year_forecast(year)
+  def initialize(year, market)
+    @forecasts = RevenueForecastService.year_forecast(year, market)
     @forecasts.sort_by! { |h| h[:project].name }
```
GabrielRMuller, line 11: I know this wasn't added in this PR, but could you add this last line in a separate method?

app/services/revenue_forecast_service.rb
```diff
@@ -48,16 +49,16 @@ def project_forecast(project, period = nil)
...
   end
 
-  def self.year_forecast(year)
-    new.year_forecast(year)
+  def self.year_forecast(year, market = nil)
+    new.year_forecast(year, market)
   end
 
-  def year_forecast(year)
+  def year_forecast(year, market = nil)
```
GabrielRMuller, line 56: do we need `market = nil` in here since we already have it on the class method? (line 52)

GabrielRMuller (approved):

app/services/revenue_forecast_service.rb
carloslopes: We don't. I'll fix this

app/presenters/revenue_forecast_presenter.rb
carloslopes: Sure!

GabrielRMuller (approved):

PR #331, opened 2022-12-21
Update SMTP configuration on staging and production environments

carloslopes:


adbatista:
Nice, but WDYT about using the same approach as Redis env?
`ENV[ENV['SMTP_SERVER']]` this way we won't have duplicate values in the HEROKU envs?

adbatista (approved):

PR #333, opened 2022-12-27
Cancel an approved vacation request

### LINKED ISSUE #332: Allow the user to cancel an approved vacation - 7 days before
Allow to a miner user to cancel his vacation until 7 days before that starts.

JonasTeixeira42:
Closes #332 

### What does this PR do?
- It allows a user to cancel an approved vacation before 7 days it started

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -11,7 +11,9 @@ class Ability
     can :read, User, id: user.id
     can %i[edit update], User, id: user.id
     can :read, Vacation, user_id: user.id
-    can :destroy, Vacation, status: :pending
+    can :destroy, Vacation do |vacation|
+      vacation[:status] == :pending || within_cancel_range?(vacation)
+    end
 
     unless user.vacations.any?(&:pending?)
       can :create, Vacation
@@ -28,5 +30,13 @@ class Ability
     if user.admin? || user.open_source_manager?
       can :read, ActiveAdmin
     end
+  end
+
+  private
+
+  MINIMUM_DAYS_TO_CANCEL = 7
+
+  def within_cancel_range?(vacation)
+    vacation[:status] == :approved && ((vacation.start_date - Date.today).to_i >= MINIMUM_DAYS_TO_CANCEL)
   end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -12,7 +12,7 @@ class Ability
     can %i[edit update], User, id: user.id
     can :read, Vacation, user_id: user.id
     can :destroy, Vacation do |vacation|
-      vacation.cancelable? vacation
+      vacation.cancelable?
     end
 
     unless user.vacations.any?(&:pending?)
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -12,7 +12,7 @@ class Ability
     can %i[edit update], User, id: user.id
     can :read, Vacation, user_id: user.id
     can :destroy, Vacation do |vacation|
-      vacation.pending? || within_cancel_range?(vacation)
+      vacation.cancelable? vacation
     end
 
     unless user.vacations.any?(&:pending?)
@@ -30,13 +30,5 @@ class Ability
     if user.admin? || user.open_source_manager?
       can :read, ActiveAdmin
     end
-  end
-
-  private
-
-  MINIMUM_DAYS_TO_CANCEL = 7
-
-  def within_cancel_range?(vacation)
-    vacation.approved? && (vacation.start_date.days_ago(MINIMUM_DAYS_TO_CANCEL) >= Date.today)
   end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -12,7 +12,7 @@ class Ability
     can %i[edit update], User, id: user.id
     can :read, Vacation, user_id: user.id
     can :destroy, Vacation do |vacation|
-      vacation[:status] == :pending || within_cancel_range?(vacation)
+      vacation.pending? || within_cancel_range?(vacation)
     end
 
     unless user.vacations.any?(&:pending?)
@@ -37,6 +37,6 @@ class Ability
   MINIMUM_DAYS_TO_CANCEL = 7
 
   def within_cancel_range?(vacation)
-    vacation[:status] == :approved && ((vacation.start_date - Date.today).to_i >= MINIMUM_DAYS_TO_CANCEL)
+    vacation.approved? && (vacation.start_date.days_ago(MINIMUM_DAYS_TO_CANCEL) >= Date.today)
   end
 end
--- a/app/models/vacation.rb
+++ b/app/models/vacation.rb
@@ -42,16 +42,16 @@ class Vacation < ApplicationRecord
     end
   end
 
-  def cancelable?(vacation)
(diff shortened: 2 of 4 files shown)


app/models/ability.rb
```diff
@@ -29,4 +31,12 @@ def initialize(user)
       can :read, ActiveAdmin
     end
   end
+
+  private
+
+  MINIMUM_DAYS_TO_CANCEL = 7
+
+  def within_cancel_range?(vacation)
+    vacation[:status] == :approved && ((vacation.start_date - Date.today).to_i >= MINIMUM_DAYS_TO_CANCEL)
```
fabiosammy, line 40: Since the vacation model is using the `enumerize` gem, you can use the helper methods that the enum add. https://github.com/brainspec/enumerize#basic

And also, you can use some helper methods from date to write a "better" way to check the dates: https://api.rubyonrails.org/classes/DateAndTime/Calculations.html#method-i-days_ago

```suggestion
    vacation.approved? && vacation.start_date.days_ago(MINIMUM_DAYS_TO_CANCEL) >= Date.today
```

app/models/ability.rb
```diff
@@ -29,4 +31,12 @@ def initialize(user)
       can :read, ActiveAdmin
     end
   end
+
+  private
+
+  MINIMUM_DAYS_TO_CANCEL = 7
+
+  def within_cancel_range?(vacation)
+    vacation.approved? && (vacation.start_date.days_ago(MINIMUM_DAYS_TO_CANCEL) >= Date.today)
```
matheusazzi, line 40: From the use case perspective, I'm thinking if we really need to check if the vacation was approved or not, maybe if it's within the cancel range that's enough and we can ditch the `vacation.approved?` check. 
cc @adbatista @JonasTeixeira42

spec/models/ability_spec.rb
```diff
@@ -25,4 +25,40 @@
...
+      let(:vacation) do
+        create(:vacation, {
+            status: :approved,
+            start_date: Date.current.next_day
+          }
+        )
+      end
+
+      it 'does not allow the user to cancel a vacation less than 7 days ahead of time' do
+        allow(Date).to receive(:today).and_return(Date.current - 5)
```
matheusazzi, line 46: Rails has some time helpers for testing code related to dates. Instead of mocking `Date` here we can use `travel_to`. You'll find examples [being used in this project](https://github.com/Codeminer42/Punchclock/search?q=travel_to).

Something like:
```ruby
it 'does not allow the user to cancel a vacation less than 7 days ahead of time' do
  create(:vacation, status: :approved, start_date: Date.tomorrow)

  travel_to 5.days.ago

  is_expected.not_to be_able_to(:destroy, vacation)
end        
```

Ref: https://api.rubyonrails.org/v5.2.4/classes/ActiveSupport/Testing/TimeHelpers.html

app/models/ability.rb
JonasTeixeira42: I think this is necessary to keep it because if we remove the status, the cancel icon will display for all vacations that are not in `pending` status

adbatista: the user can cancel in these cases
 1. vacation pending
 2. vacation approved and start date > 7 days from today

the user can not cancel in these cases
 1. Vacation canceled
 2. Vacation Rejected
 3. Vacation approved and the start date <= 7 days from today

spec/models/ability_spec.rb
```diff
@@ -25,4 +25,40 @@
...
+      end
+
+      it 'allows user the user to cancel a vacation more than 6 days ahead of time' do
+        travel_to 6.days.ago
+
+        is_expected.to be_able_to(:destroy, vacation)
+      end
+
+      it 'does not allow the user to cancel a vacation after it started' do
+        travel_to Date.current + 2.days
```
matheusazzi, line 58: TLYK we can also reach future dates using this.
```suggestion
        travel_to 2.days.from_now
```

app/models/ability.rb
```diff
@@ -29,4 +31,12 @@ def initialize(user)
       can :read, ActiveAdmin
     end
   end
+
+  private
+
+  MINIMUM_DAYS_TO_CANCEL = 7
+
+  def within_cancel_range?(vacation)
```
matheusazzi, line 39: Maybe `approved_within_cancel_range?` is more clear.

adbatista: or add Vacation#cancelable? and call here in the ability

spec/models/ability_spec.rb
JonasTeixeira42: on it!

app/models/ability.rb
JonasTeixeira42: I've made the changes, but I'm not sure if I did it right or not.  If you guys find any improvements, just let me know.

app/models/ability.rb
```diff
@@ -11,7 +11,9 @@ def initialize(user)
     can :read, User, id: user.id
     can %i[edit update], User, id: user.id
     can :read, Vacation, user_id: user.id
-    can :destroy, Vacation, status: :pending
+    can :destroy, Vacation do |vacation|
+      vacation.cancelable? vacation
```
adbatista, line 15: why is vacation being passed to itself?

JonasTeixeira42: Oops, I'm correcting it right now

app/models/vacation.rb
```diff
@@ -42,8 +42,18 @@ def deny!(user)
     end
   end
 
+  def cancelable?
+    self.pending? || approved_within_cancel_range?
```
adbatista, line 46: you don't need self here

app/models/vacation.rb
```diff
@@ -42,8 +42,18 @@ def deny!(user)
...
+  end
+
   private
 
+  MINIMUM_DAYS_TO_CANCEL = 7
+
+  def approved_within_cancel_range?
+    (self.start_date.days_ago(MINIMUM_DAYS_TO_CANCEL) >= Date.today)
```
adbatista, line 54: why the parenthesis and you also don't need self here

spec/features/admin/allocation_chart_spec.rb
```diff
@@ -3,6 +3,7 @@
 require 'rails_helper'
 
 describe 'Admin Allocation chart', type: :feature do
+  next_month_date = Date.today.next_month
```
adbatista, line 6: is not a good practice to use local variables outside examples and before, please use let instead

spec/models/ability_spec.rb
```diff
@@ -25,4 +25,40 @@
...
+  describe 'vacations abilities' do
+    describe 'when status is pending' do
+      context 'allows user to cancel a vacation' do
+        it { is_expected.to be_able_to(:destroy, Vacation.new) }
+      end
+    end
+
+    describe 'when status is approved' do
+      let!(:vacation) do
+        create(:vacation, {
```
adbatista, line 38: If the method's last argument is a hash, then you don't need braces

app/models/vacation.rb
adbatista: Just to let you know, you can also do `start_date.days_ago(MINIMUM_DAYS_TO_CANCEL).future?`

adbatista (approved):

Issue #347, opened 2023-02-24
Rake task to cancel not approved vacations


fabiosammy:
Add a rake task to cancel all the vacations not approved after the start date.

PR #348, opened 2023-02-24
Rake task to cancel not approved vacations

fabiosammy:
Solves: #347

### DIFF
--- a/app/models/vacation.rb
+++ b/app/models/vacation.rb
@@ -29,7 +29,7 @@ class Vacation < ApplicationRecord
   }
 
   scope :expired, -> {
-    where(status: :pending)
+    pending
     .where("start_date < :today", today: Date.current)
     .order(start_date: :asc, end_date: :asc)
   }
--- a/app/models/vacation.rb
+++ b/app/models/vacation.rb
@@ -28,6 +28,12 @@ class Vacation < ApplicationRecord
     .order(start_date: :asc, end_date: :asc)
   }
 
+  scope :expired, -> {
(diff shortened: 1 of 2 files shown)


lib/tasks/vacations.rake
```diff
@@ -0,0 +1,10 @@
+namespace :vacations do
+  desc "Cancel not approved expired vacations"
+  task cancel_expired: :environment do
+    Vacation.expired.each do |vacation|
+      puts "Canceling #{vacation.user.name}'s expired vacation"
+      vacation.status = :denied
```
adbatista, line 6: since the task is to cancel, is better to use `:cancelled`status

lib/tasks/vacations.rake
```diff
@@ -0,0 +1,10 @@
+namespace :vacations do
+  desc "Cancel not approved expired vacations"
+  task cancel_expired: :environment do
+    Vacation.expired.each do |vacation|
+      puts "Canceling #{vacation.user.name}'s expired vacation"
```
adbatista, line 5: since this will run on a cron, we need to use Rails.logger.info

app/models/vacation.rb
```diff
@@ -28,6 +28,12 @@ class Vacation < ApplicationRecord
     .order(start_date: :asc, end_date: :asc)
   }
 
+  scope :expired, -> {
+    where(status: :pending)
```
adbatista, line 32: just to let you know, the enumerize has the scope set so you can just call `.pending` here

Issue #353, opened 2023-02-28
Show user's mentee in the admin 


fabiosammy:
List all mentees of a user on the admin page.

PR #354, opened 2023-02-28
feat: add mentees row in user's admin page

GabrielRMuller:
Addresses #353 

image

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -99,6 +99,7 @@ ActiveAdmin.register User do
           end
           row :skills
           row :mentor
+          row :mentees
           row :allow_overtime
           row :active
           row :started_at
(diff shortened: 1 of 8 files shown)


app/admin/user.rb
```diff
@@ -99,6 +99,9 @@
           end
           row :skills
           row :mentor
+          row :mentees do |user|
+            User.mentees_for(user)
```
adbatista, line 103: why not a relation and call: `user.mentees` ?

GabrielRMuller: My bad, I assumed we didn't wanted to interfere with any relations. Just updated the PR to use a `has_many` relationship instead.

adbatista (approved):

Issue #355, opened 2023-03-06
Do not allow a vacation request close to a HOLIDAY


fabiosammy:
Do not allow a normal user to request a vacation if the start day is 2 days close to the holiday

fabiosammy:
@Isaac-alencar We had a PR already, thanks!

PR #356, opened 2023-03-06
Do not allow a vacation request close to a HOLIDAY

fabiosammy:
Solves: #355


app/models/vacation.rb
```diff
@@ -25,6 +26,7 @@ class Vacation < ApplicationRecord
     allow_nil: true,
     if: :not_cancelled?
   validate :validate_start_date_close_to_weekend, if: :start_date
+  validate :validate_start_date_close_to_holiday, if: :start_date
```
GabrielRMuller, line 29: maybe we can consolidate both of these validations into one, we could call it `validate :start_date_close_to_off_day`

GabrielRMuller: also, I think it would be nice if we moved all this logic into a separate validator, it would keep this model a bit cleaner :eyes:

spec/models/vacation_spec.rb
```diff
@@ -104,6 +108,71 @@
...
+      end
+
+      before do
+        allow(vacation.user).to receive(:office_holidays).and_return([{day: 01, month: 01}])
+      end
+
+      it { expect(vacation).to_not be_valid }
+    end
+
+    context 'when the start_date is after of a holiday' do
```
GabrielRMuller, line 161:
```suggestion
    context 'when the start_date is after a holiday' do
```

GabrielRMuller (dismissed):
bulky but reliable tests, nice! :+1:

app/models/vacation.rb
```diff
@@ -86,7 +88,20 @@ def not_cancelled?
...
+
+  def validate_start_date_close_to_holiday
+    return unless ENV['VALIDATE_VACATION_BEFORE_HOLIDAY'] == 'true'
+    if (user.office_holidays & restrict_dates).any?
+      errors.add(:start_date, :close_holiday)
+    end
+  end
+
+  def restrict_dates
+    (start_date..(start_date + MINIMUM_DAYS_BEFORE_HOLIDAY.days)).to_a.map do |date|
```
Isaac-alencar, line 103: I'm not quite sure, but I think this was misunderstood. According to CLT,  _§ 3 It is forbidden to start vacations in the period of two days before a public holiday or paid weekly rest day_
So, I think we would to like to prevent schedule vacations 2 days before a holiday (and the holiday itself) 

my suggestion to fix it:
```ruby
((start_date - MINIMUM_DAYS_BEFORE_HOLIDAY.days)..start_date).to_a
```

adbatista (dismissed):

GabrielRMuller (commented):
I'd really like if we moved the entire vacation/holiday logic into a separate [custom validator](https://guides.rubyonrails.org/active_record_validations.html#custom-validators), I think this domain logic is complex and big enough to grant it's own class, and it'd be helpful to move to its own separate place.

adbatista (approved):

PR #362, opened 2023-03-16
feat: allow admin to resend the user reg email through admin

GabrielRMuller:
Adds link on the User's admin show page that allows to resend the user's registration email.

imageimage

### DIFF
--- a/app/mailers/notification_mailer.rb
+++ b/app/mailers/notification_mailer.rb
@@ -12,6 +12,12 @@ class NotificationMailer < ActionMailer::Base
     mail(to: @user.email, subject: 'Welcome to Punchclock')
   end
 
+  def resend_user_registration(user, token)
+    @user = user
+    @token = token
+    mail(to: @user.email, subject: 'Welcome to Punchclock', template_name: 'notify_user_registration')
+  end
+
   def notify_user_password_change(user, new_password)
     @user = user
     @new_password = new_password
--- a/spec/mailers/notification_mailer_spec.rb
+++ b/spec/mailers/notification_mailer_spec.rb
@@ -220,5 +220,34 @@ describe NotificationMailer do
         expect(mail.body).to match(contributions)
       end
     end
+
+    context 'when a registration email is sent again through admin' do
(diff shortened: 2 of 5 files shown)


app/mailers/notification_mailer.rb
```diff
@@ -12,6 +12,13 @@ def notify_user_registration(user)
     mail(to: @user.email, subject: 'Welcome to Punchclock')
   end
 
+  def resend_user_registration(user)
+    @user = user
+    user.touch(:confirmed_at)
+    @token = user.send(:set_reset_password_token)
+    mail(to: @user. email, subject: 'Welcome to Punchclock', template_name: 'notify_user_registration')
```
GabrielRMuller, line 19: thoughts on using the same template from `notify_user_registration`? should I add a new one even though they're identical? (so this method is not dependent on how `notify_user_registration` looks).

app/mailers/notification_mailer.rb
```diff
@@ -12,6 +12,13 @@ def notify_user_registration(user)
     mail(to: @user.email, subject: 'Welcome to Punchclock')
   end
 
+  def resend_user_registration(user)
+    @user = user
+    user.touch(:confirmed_at)
+    @token = user.send(:set_reset_password_token)
```
GabrielRMuller, line 18: Do you feel a transaction block is necessary here?

fabiosammy: Do we need the touch?

GabrielRMuller: I think so yeah, otherwise the reset password request fails because this field is empty.

spec/mailers/notification_mailer_spec.rb
```diff
@@ -220,5 +220,34 @@
         expect(mail.body).to match(contributions)
       end
     end
+
+    context 'when a registration email is sent again through admin' do
+      let(:user) { create(:user, name: "Art Vandelay") }
```
GabrielRMuller, line 225: on the test environment the mailer does not like names with any form of accentuation such as `´`, `^`, etc.

Issue #361, opened 2023-03-16
New admin page for allocation chart


samkevin1:
Continuing the rails active admin refactor, let's create a new admin page for the allocation chart page.

PR #363, opened 2023-03-16
feat: New admin allocation chart

samkevin1:
Issue: #361 

image


app/views/new_admin/allocation_chart/_allocations.html.erb
```diff
@@ -0,0 +1,76 @@
...
+    <th class="px-4 py-3"><%= I18n.t('english_evaluation') %></th>
+    <th class="px-4 py-3"><%= I18n.t('skills') %></th>
+  </tr>
+  </thead>
+  <tbody class="bg-white divide-y dark:divide-gray-700 dark:bg-gray-800">
+  <% allocations.each do |allocation| %>
+    <tr class="text-gray-700 dark:text-gray-400">
+      <td class="px-4 py-3">
+        <div class="flex items-center text-sm gap-1">
+          <%= gravatar_for allocation[:user], { size: 30, css_class: 'object-cover rounded-full' } %>
```
adbatista, line 18: there's no need for this in this page

app/controllers/new_admin/allocation_chart_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      @allocations = allocations.reduce([]) do |acc, allo|
+        allocation = { :name => allo.user.first_and_last_name,
+                       :project_name => allo.project ? allo.project.name : nil,
+                       :level => decorated_user(allo).level,
+                       :specialty => decorated_user(allo).specialty,
+                       :english_evaluation => decorated_user(allo).english_level,
+                       :skills => decorated_user(allo).skills,
+                       :allocation_end_at => build_date(allo.end_at),
+                       :user => allo.user }
```
adbatista, line 18: Why change the allocation to a hash?

adbatista (changes requested):

app/views/application/_sidebar.html.erb
```diff
@@ -14,6 +14,16 @@
...
         <% end %>
       </li>
+      <li class="relative px-6 py-3">
+        <%= link_to new_admin_allocation_chart_index_path, class: "inline-flex items-center w-full text-sm font-semibold transition-colors duration-150 hover:text-gray-800 dark:hover:text-gray-200" do %>
+          <svg class="w-6 h-6" aria-hidden="true" fill="none" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" viewBox="0 0 24 24" stroke="currentColor">
+            <path d="M12 14l9-5-9-5-9 5 9 5z"></path>
+            <path d="M12 14l6.16-3.422a12.083 12.083 0 01.665 6.479A11.952 11.952 0 0012 20.055a11.952 11.952 0 00-6.824-2.998 12.078 12.078 0 01.665-6.479L12 14z"></path>
+            <path stroke-linecap="round" stroke-linejoin="round" d="M12 14l9-5-9-5-9 5 9 5zm0 0l6.16-3.422a12.083 12.083 0 01.665 6.479A11.952 11.952 0 0012 20.055a11.952 11.952 0 00-6.824-2.998 12.078 12.078 0 01.665-6.479L12 14zm-4 6v-7.5l4-2.222"></path>
+          </svg>
+          <span class="ml-4">Organograma</span>
```
GabrielRMuller, line 24: `I18n` for 'Organograma'?

app/controllers/new_admin/allocation_chart_controller.rb
samkevin1: `@allocations` is missing some fields so I formatted it into a hash. Is there a better way of doing this?

app/views/new_admin/allocation_chart/_allocations.html.erb
```diff
@@ -0,0 +1,76 @@
...
+          <p class="text-sm text-gray-600 dark:text-gray-400">
+            <%= allocation[:allocation_end_at] %>
+          </p>
+        </div>
+        </div>
+      </td>
+      <td class="px-4 py-3">
+        <div class="flex items-center text-sm">
+          <p class="text-sm text-gray-600 dark:text-gray-400">
+            <%= allocation[:level].capitalize if allocation[:level] %>
```
GabrielRMuller, line 44: doesn't the decorator handle empty values already? Also, if `level` is set as `'N/A'`, `capitalize` will turn into `'N/a`

https://github.com/Codeminer42/Punchclock/blob/0dba9a9bf4011fbb236769d42ec5f18ce573e446/app/decorators/user_decorator.rb#L10-L12

app/controllers/new_admin/allocation_chart_controller.rb
samkevin1: changed it

app/controllers/new_admin/allocation_chart_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module NewAdmin
+  class AllocationChartController < ApplicationController
+    layout "new_admin"
+
+    def index
+      allocations = AllocationsAndUnalocatedUsersQuery.new(Allocation).call
+
+      @allocations = allocations.map { |allocation| decorated_allocation(allocation) }
```
adbatista, line 10: you can just call allocations.decorate

samkevin1: you mean like this? 
```rb
@allocations = allocations.decorate
```

it doesn't seem to work

fabiosammy: The right syntax is to use the collection decorate. @samkevin1, pay attention that you're using the draper to decorate the model. If you do not know about it, here is the documentation :) :

https://github.com/drapergem/draper#collections

samkevin1: Yeah I wasn't understanding how to use it very well, but I changed it

app/decorators/allocation_decorator.rb
```diff
@@ -10,4 +10,42 @@ def hourly_rate
   def to_s
     "##{model.id}"
   end
+
+  def user_name
+    model.user.first_and_last_name
+  end
+
+  def project_name
+    model.project ? model.project.name : nil
```
adbatista, line 19: Instead, you could use delegate in the model. https://api.rubyonrails.org/classes/Module.html#method-i-delegate

adbatista:
the specs are missing

feliperli (approved):

spec/controllers/new_admin/allocation_chart_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        expect(response).to have_http_status(:ok)
+      end
+
+      it 'returns users within an empty allocation' do
+        get :index
+
+        allocations = subject.instance_variable_get(:@allocations).select { |allocation| allocation.user_id == user.id }
+        user_id = allocations[0].user_id
+
+        expect(user_id).to eq(user.id)
```
GabrielRMuller, line 27: instead of doing a `select` first and then checking if it's equal, you can just use a `include` matcher: https://www.rubydoc.info/gems/rspec-expectations/RSpec%2FMatchers:include

spec/features/admin/allocation_chart_spec.rb
```diff
@@ -47,8 +47,10 @@
         end
 
         it '"Alocado até" column links to allocation' do
+          allow(Date).to receive(:today).and_return(Date.new(2023, 4, 1))
+
           within 'tbody' do
-            expect(page).to have_link('01/11/2022', href: "/admin/allocations/#{allocation.id}")
+            expect(page).to have_link('01/03/2023', href: "/admin/allocations/#{allocation.id}")
```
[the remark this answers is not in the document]
samkevin1: After 61 days of the allocation's end date the user will not be shown as `allocated` anymore, so the link that this test is trying to test will not exist. This test will break again in the future because of the date

spec/controllers/new_admin/allocation_chart_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      it 'returns successful status' do
+        get :index
+
+        expect(response).to have_http_status(:ok)
+      end
+
+      it 'returns users within an empty allocation' do
+        get :index
+
+        expect(subject.instance_variable_get(:@allocations)).to include(having_attributes(user_id: user.id))
```
fabiosammy, line 24: What are you expecting here is reflecting what is described in the test? And I'm curious, why are you using the `instance_variable_get`?

app/views/application/_sidebar.html.erb
```diff
@@ -14,6 +14,14 @@
           <span class="ml-4">Mentoring</span>
         <% end %>
       </li>
+      <li class="relative px-6 py-3">
+        <%= link_to new_admin_allocation_chart_index_path, class: "inline-flex items-center w-full text-sm font-semibold transition-colors duration-150 hover:text-gray-800 dark:hover:text-gray-200" do %>
+          <svg class="w-6 h-6" data-darkreader-inline-stroke="" fill="none" stroke="currentColor" stroke-width="1.5" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
+            <path stroke-linecap="round" stroke-linejoin="round" d="M7.5 14.25v2.25m3-4.5v4.5m3-6.75v6.75m3-9v9M6 20.25h12A2.25 2.25 0 0020.25 18V6A2.25 2.25 0 0018 3.75H6A2.25 2.25 0 003.75 6v12A2.25 2.25 0 006 20.25z"></path>
+          </svg>
```
nolleto, line 21: @adbatista just checking with you first... would not be better to save this `<svg>` in a file `.svg` and use it like:

```ruby
 <%= image_tag '[image-name].svg', class: 'w-6 h-6' %>
```
?

adbatista: 🤔 Yeah makes sense, this way we keep things separated

spec/features/admin/allocation_chart_spec.rb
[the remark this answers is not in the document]
adbatista: Where is this 61 days rule?

spec/controllers/new_admin/allocation_chart_controller_spec.rb
samkevin1: It is, the allocation object doesn't have an ID or any other fields besides the `user_id` so I called it an empty allocation.

I'm trying to get the `@allocations` variable, is there a better way of doing this?

app/views/application/_sidebar.html.erb
nolleto: @samkevin1 can you do that?

samkevin1: sure, i'll change it

spec/features/admin/allocation_chart_spec.rb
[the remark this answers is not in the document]
samkevin1: `allocation_chart_helper.rb`

app/views/application/_sidebar.html.erb
samkevin1: in which folder should I save these `.svg`s?

adbatista: app/assets/images

samkevin1: when I use `image_tag` it becomes a  tag and I can't change the color of the icon, any suggestions of what I can do to fix this?

nolleto: Well... I think you have 3 options:

A) Create our own helper to render the SVG as `<svg />`: https://stackoverflow.com/a/36992492/4538055
B) Use this gem: [inline_svg](https://github.com/jamesmartin/inline_svg)
C) Keep it the way it is.

I didn't realize `image_tag` would create a `` tag. I prefer option `B` over option `A` just to make sure it works  but maybe we don't want another gem and option `A` can get complicated (maybe)

So, we would go with option `C`.

Maybe @adbatista can help us with that question.

samkevin1: Forgot to comment, but I changed this. I'm saving the svg's in `application/images` and rendering it on the page. Can you take a look again?

samkevin1: Changed it, it's using `inline_svg` now @adbatista @nolleto

nolleto (approved):

spec/controllers/new_admin/allocation_chart_controller_spec.rb
adbatista: instead of verifying the instance variable, you could verify if the content of allocations is displayed in the rendered page

Issue #371, opened 2023-03-31
Hide the resend e-mail button in the admin page when a user has already confirmated


fabiosammy:
![image](https://user-images.githubusercontent.com/972184/229113328-7c82e3f0-6d12-405d-8b65-77c856557f16.png)

PR #377, opened 2023-04-05
Fix: only show resend confirmation email button when user is not conf…

luccamps:
Solves: #371

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -55,8 +55,8 @@ ActiveAdmin.register User do
    link_to I18n.t('hour_report_past_month'), hour_report_admin_users_path(format: :xlsx, month: :past)
   end
 
-  action_item :user_registration, only: :show, priority: 0 do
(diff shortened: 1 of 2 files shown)


app/admin/user.rb
```diff
@@ -56,7 +56,7 @@
   end
 
   action_item :user_registration, only: :show, priority: 0 do
-    link_to I18n.t('resend_user_registration'), resend_user_registration_admin_user_path, method: :patch
+    link_to I18n.t('resend_user_registration'), resend_user_registration_admin_user_path, method: :patch unless user.confirmed?
```
GabrielRMuller, line 59: instead of using `unless` in here you can pass it as an option on `action_item` instead, check this example on `admin/questionnaire.rb`

https://github.com/Codeminer42/Punchclock/blob/21341156199ea8a8e78da25c85f5232888ee9f56/app/admin/questionnaires.rb#L18-L21

GabrielRMuller (commented):
just a small adjustment, looks good otherwise :+1:

GabrielRMuller (approved):

adbatista (approved):

Issue #346, opened 2023-02-24
WIP: Change how the regional holidays works


fabiosammy:
Today, the holidays are added to an office, but some users do not follow those regional holidays, and they have their own holidays.

Not sure how to proceed on it, maybe adding a new feature to a user skip the holiday and add their own(by HR user), or change the relation (need to check the effort that will be necessary for a national holiday)

PR #394, opened 2023-04-19
feat: regional holidays using cities

luccamps:
WIP: #346 

After data is normalized we can remove offices_regional_holidays relationship.

### DIFF
--- a/app/admin/regional_holiday.rb
+++ b/app/admin/regional_holiday.rb
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 ActiveAdmin.register RegionalHoliday do
-  permit_params :name, :day, :month, office_ids: []
+  permit_params :name, :day, :month, office_ids: [], city_ids: []
 
   filter :name
   filter :offices, multiple: true, collection: -> { Office.active.order(:city) }
+  filter :cities, multiple: true, collection: -> { City.all.order(:name) }
   filter :day
   filter :month
 
@@ -14,6 +15,9 @@ ActiveAdmin.register RegionalHoliday do
     column :month
     column :offices do |holiday|
       offices_by_holiday(holiday)
+    end
+    column :cities do |holiday|
+      holiday.cities
     end
     actions
   end
@@ -27,6 +31,9 @@ ActiveAdmin.register RegionalHoliday do
       row :offices do |holiday|
         offices_by_holiday(holiday)
       end
+      row :cities do |holiday|
+        holiday.cities
+      end
     end
   end
 
@@ -36,6 +43,7 @@ ActiveAdmin.register RegionalHoliday do
       f.input :day
       f.input :month
       f.input :offices, as: :check_boxes, collection: Office.active.order(:city)
+      f.input :cities, as: :check_boxes, collection: City.all.order(:name)
     end
       f.actions
   end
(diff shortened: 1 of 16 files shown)


app/admin/regional_holiday.rb
```diff
@@ -15,6 +16,9 @@
     column :offices do |holiday|
       offices_by_holiday(holiday)
     end
+    column :cities do |holiday|
```
adbatista, line 19: `column :cities` does not work?

app/admin/regional_holiday.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 ActiveAdmin.register RegionalHoliday do
-  permit_params :name, :day, :month, office_ids: []
+  permit_params :name, :day, :month, office_ids: [], city_ids: []
 
   filter :name
   filter :offices, multiple: true, collection: -> { Office.active.order(:city) }
+  filter :cities, multiple: true, collection: -> { City.all.order(:name) }
```
adbatista, line 8: you can do `City.order(:name)`

app/admin/regional_holiday.rb
```diff
@@ -36,6 +43,7 @@
       f.input :day
       f.input :month
       f.input :offices, as: :check_boxes, collection: Office.active.order(:city)
+      f.input :cities, as: :check_boxes, collection: City.all.order(:name)
```
adbatista, line 46: this might list a lot of cities, maybe is better to use a select box

spec/factories/cities.rb
```diff
@@ -2,5 +2,11 @@
   factory :city do
     sequence(:name) {|i| "City Name #{i}" }
     state { create(:state) }
+
+    trait :with_holiday do
+      after :create do |city|
+        city.regional_holidays = create_list(:regional_holiday, 5)
```
adbatista, line 8: is this really creating the association in the db?

spec/models/city_spec.rb
```diff
@@ -3,9 +3,31 @@
...
   context 'Validations' do
     it { is_expected.to validate_presence_of(:name) }
   end
+
+  describe '#holidays' do
+    let!(:city) { create(:city) }
+
+    context 'when there are no holidays' do
+      it 'returns an empty array' do
+        expect(city.holidays).to eq([])
```
adbatista, line 18:
```suggestion
        expect(city.holidays).to be_empty
```

app/admin/regional_holiday.rb
luccamps: It does. It's fixed now.

app/admin/regional_holiday.rb
luccamps: I didn't know it was possible to do it that way. It's fixed now!

app/admin/regional_holiday.rb
luccamps: Done!

image

spec/factories/cities.rb
luccamps: It does. If I debug it and execute `city.regional_holidays` it returns all holidays associated with the city. I used the same trait on offices as inspiration.

adbatista: ok, just one last check, what happens if you do `city.reload.regional_holidays`

luccamps: It shows the associated cities. For a moment I thought I would have to manually save with `city.save` but it wasn't required. I've also tried `City.find(city.id).regional_holidays` and got the holidays as well. It seems like it's indeed persisted.

image

adbatista: nice

adbatista (approved):

Issue #399, opened 2023-04-25
Add Finished Vacations Scope


luccamps:
![image](https://user-images.githubusercontent.com/40892570/234367656-75f8465f-2411-489d-bc65-fdec5c4f8f97.png)

In `/admin/vacations`, add a "Finished" scope that lists approved vacations that have been concluded already starting by the most recent.

PR #400, opened 2023-04-26
feat: add finished scope to vacation

luccamps:
Solves #399

### DIFF
--- a/app/models/vacation.rb
+++ b/app/models/vacation.rb
@@ -60,6 +60,12 @@ class Vacation < ApplicationRecord
     where("#{approver}_approver_id": nil)
   }
 
+  scope :finished, -> {
+    where(status: :approved)
+    .where("end_date <= :today", today: Date.current)
+    .order(end_date: :desc)
+  }
+
   def approve!(user)
     ActiveRecord::Base.transaction do
       update!(hr_approver: user) if user.hr?
--- a/app/models/vacation.rb
+++ b/app/models/vacation.rb
@@ -61,7 +61,7 @@ class Vacation < ApplicationRecord
   }
 
   scope :finished, -> {
-    where(status: :approved)
+    approved
     .where("end_date <= :today", today: Date.current)
     .order(end_date: :desc)
   }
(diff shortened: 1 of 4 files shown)


app/models/vacation.rb
```diff
@@ -60,6 +60,12 @@ class Vacation < ApplicationRecord
     where("#{approver}_approver_id": nil)
   }
 
+  scope :finished, -> {
+    where(status: :approved)
```
adbatista, line 64: what about using `approved` instead this where

spec/models/vacation_spec.rb
```diff
@@ -151,6 +151,33 @@
...
+      expect(described_class.finished).to eq(ended_vacations)
+    end
+
+    it 'returns finished vacations sorted by most recent' do
+      ended_vacation.update(end_date: 2.days.ago)
+      sorted_ended_vacations = ended_vacations.push(ended_vacation)
+
+      expect(described_class.finished).to eq(sorted_ended_vacations)
+    end
+  end
```
adbatista, line 179:
```suggestion
  describe '.finished' do
    let!(:ended_vacation1) { create(:vacation, :ended, end_date: 1.day.ago) }
    let!(:ended_vacation2) { create(:vacation, :ended, end_date: 2.days.ago) }
    let!(:ended_vacation3) { create(:vacation, :ended, end_date: 3.days.ago) }

    before do
      create(:vacation, :approved)
      create(:vacation, :pending)
      create(:vacation, :cancelled)
      create(:vacation, :denied)
      create(:vacation, :ongoing)
      create(:vacation, :scheduled)
      create(:vacation, :expired)
      create(:vacation, :valid)
    end
    
    it 'returns only finished vacations sorted by most recent' do
      expect(described_class.finished).to eq([ended_vacation1, ended_vacation2, ended_vacation3])
    end
  end
```

app/models/vacation.rb
luccamps: Way better, totally missed that example that used pending.

spec/models/vacation_spec.rb
luccamps: I thought that would look ugly but between two uglies better pick the lesser ugly :^). I have also removed that redundant test.

adbatista (approved):

adbatista (approved):

Issue #411, opened 2023-05-16
allow admin to add user experience on event talks


adbatista:
An admin should be able to add the experience of the user talking at events to the user profile
a user can have none, one or more experiences
it should have three fields

Event name, talk title, and date

PR #422, opened 2023-05-25
Feat - Allow admin to add user experience on event talks

JonasTeixeira42:
This PR solves the issue #411 

- It creates the new talking experience register page for active admin
image


app/models/talking.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class Talking < ApplicationRecord
```
adbatista, line 3: Try to add the model name of a resource instead an action; actions should be the name of functions or commands.

A proper name here would be `Talk` or Event

app/models/talking.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  validates :event_name, :talk_title, :date, presence: true
+  validate :future_date?
+
+  private
+
+  def future_date?
+    return if date.blank?
+
+    errors.add(:date, "must be before the today's date") if date.future?
```
adbatista, line 14: Add the translation in the i18n here; active record already provides an approach for this

spec/admin/talking_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Talking do
```
adbatista, line 5: nice

app/models/talking.rb
JonasTeixeira42: on it

app/models/talk.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  validates :event_name, :talk_title, :date, presence: true
+  validate :future_date?
+
+  private
+
+  def future_date?
+    return if date.blank?
+
+    errors.add(:date, I18n.t('activerecord.errors.models.talk.attributes.date.future_date')) if date.future?
```
adbatista, line 14: use a type instead adding the call to i18n here, Active record already do this https://api.rubyonrails.org/classes/ActiveModel/Errors.html#method-i-add

JonasTeixeira42: on it

adbatista (approved):

PR #431, opened 2023-05-30
Change users to be fullstack

carloslopes:


adbatista (approved):

app/decorators/user_decorator.rb
```diff
@@ -7,6 +7,18 @@ def current_allocation
     super || ENV["DEFAULT_PROJECT_NAME"] || I18n.t('not_allocated')
   end
 
+  def frontend_level
+    return 'N/A' unless model.frontend_level
+
+    "Frontend #{model.frontend_level.humanize}"
```
talyssonoc, line 13: What do you think about not prepending it with "Frontend" or "Backend"? Leaving it just as:

```suggestion
    model.frontend_level.humanize"
```

carloslopes: As we talked on Discord, since the table header is fixed on the top, I used this approach to improve the readability

PR #432, opened 2023-05-31
[FEAT-414] Allow user to add the PR description

JonasTeixeira42:
This PR solves issue #414

- It creates a new page so that the user can add the PR description 
image
image


app/assets/builds/tailwind.css
adbatista, line 1: this file can be removed

app/models/contribution.rb
```diff
@@ -64,4 +66,10 @@ def update_rejected_reason(rejected_reason = "other_reason", reviewer_id)
   scope :valid_pull_requests, -> { where.not(state: :refused) }
   scope :without_pr_state, ->(state) { where.not(pr_state: state) }
   scope :tracking, -> { where(tracking: true) }
+
+  private
+
+  def normalize_description_blank_values
+    self[:description] = description.blank? ? nil : description
```
adbatista, line 73: you can do here `self[:description] = description.presence`

spec/controllers/contributions_controller_spec.rb
```diff
@@ -45,9 +47,103 @@
...
+    end
+  end
+
+  describe 'PUT update' do
+    let(:contribution) { create(:contribution) }
+    let(:params) do
+      {
+        id: contribution.id,
+        contribution: {
+          description:
```
adbatista, line 92: add the `description: description` just to be consistent with the other parts of the code

app/models/contribution.rb
JonasTeixeira42: on it

app/views/contributions/edit.html.erb
```diff
@@ -0,0 +1,9 @@
+<ul class="breadcrumb">
+  <li class="breadcrumb-item"><%= link_to t(:back, scope: :helpers), contributions_path %></li>
+</ul>
+
+<h1>
+  <%= t(:editing, scope: %i(helpers actions), model: Contribution.model_name.human) %>
```
adbatista, line 6: here should be something like "Add description" or something like

adbatista (approved):

PR #443, opened 2023-06-02
Add management pages to User Skills

carloslopes:


adbatista (approved):
just check/remove if still exists the input of the skill in the user form

carloslopes:
good catch! I will make the change there

PR #451, opened 2023-06-15
feat: export user resume as .docx

### LINKED ISSUE #415: allow admin to export the user experiences in docx format
gem to use https://github.com/ruby-docx/docx

gabrielcosta42:
Generates a .docx document to export the user experiences tab

Adding 2 .docx to the codebase:
- template.docx is the structure of the final docment
- template-elements.docx which holds the elements to mount the doc

closes #415


app/admin/user.rb
```diff
@@ -335,4 +339,28 @@ def create
...
+    doc.mount_professional_section(job_experiences)
+
+    education_experiences = user.education_experiences.decorate
+    doc.mount_education_section(education_experiences)
+
+    contribution_experiences = user.contributions.approved.order(created_at: :asc).decorate
+    doc.mount_contribution_section(contribution_experiences)
+
+    talk_experiences = user.talks.decorate
+    doc.mount_talk_section(talk_experiences)
```
adbatista, line 362: this could be in a external class

app/decorators/professional_experience_decorator.rb
```diff
@@ -8,6 +8,17 @@ def start_date
   end
 
   def end_date
-    model.start_date.to_date.to_fs(:date)
+    return I18n.t(:present) if model.end_date.nil?
+    model.end_date.to_date.to_fs(:date)
+  end
+
+  def start_date_month
+    return I18n.t(:present) if model.start_date.nil?
```
adbatista, line 16: if has no start date it did not start, so should not have this record

gabrielcosta42: that's true. solved.

app/admin/user.rb
gabrielcosta42: should it also go in the `UserResumeDocService`?

adbatista: no, UserResumeDoc, it should not be a service, but ok.
User resume doc describes the word doc, and we should have a class that builds the doc

gabrielcosta42: done

Gemfile
```diff
@@ -38,14 +38,16 @@ gem 'kaminari'
...
+
 gem 'httparty'
 gem 'github_api', '~> 0.18.2'
 gem 'active_model_serializers', '~> 0.10.13'
 gem 'money-rails', '~> 1.12'
 
 gem 'ransack', '~> 2.3'
 
-gem 'sidekiq', "< 7"
+gem 'sidekiq', "< 8"
```
adbatista, line 50: this is not necessary now and redis on heroku does not support yet

adbatista (approved):

Gemfile
gabrielcosta42: ok, removing it.

Issue #450, opened 2023-06-14
`/contributions` page returns HTTP 500 when not authenticated


danielmbrasil:
### Steps to reproduce the issue
Try accessing the `/contributions` page without being logged-in

### Expected behavior
Redirects to the sign-in page? But not an error

### Actual behavior
HTTP 500
![image](https://github.com/Codeminer42/Punchclock/assets/51726526/f235f99b-68ac-4c17-82fe-07b5382f25f2)

PR #456, opened 2023-07-04
Redirect unauthenticated users to the sign in page

Isaac-alencar:
# Fix: `/contributions` page returns HTTP 500 when not authenticated (#450 )

This PR adds a `before_action authenticate_user` in `contributions_controller` to redirect users to sign in page if they're note logged in.

Also, I changed the spec file by dividing into two contexts: `users logged in` and `users not logged in`


adbatista (approved):

Isaac-alencar:
New link: https://github.com/Codeminer42/Punchclock/pull/457

PR #457, opened 2023-07-04
Redirect unauthenticated users to the sign in page

Isaac-alencar:
# Fix: `/contributions` page returns HTTP 500 when not authenticated (#450 )

This PR adds a `before_action authenticate_user` in `contributions_controller` to redirect users to sign in page if they're note logged in.

Also, I changed the spec file by dividing into two contexts: `users logged in` and `users not logged in`


adbatista (approved):

spec/controllers/contributions_controller_spec.rb
```diff
@@ -10,140 +10,173 @@
...
-      get :index
-
-      expect(response).to render_template(:index)
-    end
-
-    it 'has status 200' do
-      get :index
-
-      expect(response).to have_http_status(:ok)
+  context 'when the user are logged in' do
```
danielmbrasil, line 13:
```suggestion
  context 'when the user is logged in' do
```

spec/controllers/contributions_controller_spec.rb
```diff
@@ -10,140 +10,173 @@
...
 
-        expect(response).to have_http_status(:ok)
+      before do
+        allow(user.contributions).to receive(:find).with(contribution.id.to_s) { contribution }
       end
 
-      it 'is expected to render the edit page' do
-        get :edit, params: params
+      context 'when the page is accessed' do
+        it 'is expected to return the successful status' do
```
danielmbrasil, line 69:
```suggestion
        it 'is expected to return a successful status' do
```

spec/controllers/contributions_controller_spec.rb
```diff
@@ -10,140 +10,173 @@
...
+              )
+            )
         end
       end
     end
+  end
 
-    context 'when the parameters are valid' do
-      let(:description) { 'description' }
+  context 'when the user are not logged in' do
```
danielmbrasil, line 153:
```suggestion
  context 'when the user is not logged in' do
```

danielmbrasil (commented):
A few suggestions.

PR #459, opened 2023-07-05
Update Capybara setup

carloslopes:
The old setup was causing an ArgumentError


talyssonoc (approved):
Assuming we don't need those capabilities anymore, LGTM

adbatista (approved):

Issue #467, opened 2023-07-14
Redesign Open Source page adding "highlight/recomendado" tag


joaoGabriel55:
I think that would be good provide a new design to Open Source page and showing "highlight/recomendado" tag

- Depends of issue #463

![image](https://github.com/Codeminer42/Punchclock/assets/23138717/6a39bdc1-8ea2-4229-a204-8ddd059f628a)

[Figma link](https://www.figma.com/file/czN2pN2w733E1fEckB2niW/Untitled?type=design&node-id=0%3A1&mode=design&t=lA3MbgF2XpHBNFQ9-1)

### Requirements

- To retrieve Issue, PR and Star numbers we can use Github API to show these infos
   https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository

mwilding-godaddy:
![253591335-6a39bdc1-8ea2-4229-a204-8ddd059f628a](https://github.com/Codeminer42/Punchclock/assets/125277981/d4b8bea1-892e-494a-b7bf-91e7f6cf8039)

What do you think of moving the "Recomendado" tag to bellow the forks and stars?

adbatista:
nice, I'll add the tag hold, because I'm looking maybe to change the theme

PR #479, opened 2023-07-18
[Frontend]: Redesign Open Source page adding "highlight/recomendado" tag

joaoGabriel55:
Closes Issue: #467 

Updated preview
image


app/services/github/repositories/sync.rb
```diff
@@ -32,6 +27,33 @@ def call
...
+      end
+
+      def update_repository_languages(repository)
+        repository_owner, repository_name = repository_owner_and_name(repository)
+        request = client.repos.languages(repository_owner, repository_name)
+
+        if request.success?
+          languages = request.body.keys[0..MAX_LANGUAGES_NUMBER-1].join(', ')
+
+          repository.update(language: languages)
```
thiagofportella, line 42: Maybe it's a good idea to use the `bang!` method here.
```suggestion
          repository.update!(language: languages)
```

https://github.com/rubocop/rails-style-guide#save-bang

thiagofportella (approved):

app/services/github/repositories/sync.rb
joaoGabriel55: Thanks for the tip!

spec/features/repositories_spec.rb
```diff
@@ -63,9 +73,76 @@
...
+    end
+
+    it 'have all stars for repositories' do
+      expect(page).not_to have_css('#stars-tag')
+    end
+  end
+
+  context 'when the repository has issues' do
+    context 'when issues are greater than 1000' do
+      let!(:repository) { create(:repository, issues: '5455').decorate }
```
joaoGabriel55, line 99: Here, you can create a `repository` with issues as a integer. Like that:
```rb
let!(:repository) { create(:repository, issues: 5455).decorate }
```

spec/features/repositories_spec.rb
```diff
@@ -63,9 +73,76 @@
...
+
+      before do
```
joaoGabriel55, line 101: I think that you could put this code bloc between lines 97 and 98, to avoid repeat every time

spec/features/repositories_spec.rb
joaoGabriel55, line 99: Here, you can create a `repository` with issues as a integer. Like that:
```rb
let!(:repository) { create(:repository, issues: 5455).decorate }
```

spec/features/repositories_spec.rb
Guilherme-NL: If I try to move this code bloc to the upper context, tests will fail. Because I will define this code bloc before to define the repository

app/assets/images/issues.svg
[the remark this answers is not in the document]
joaoGabriel55: I created them using Figma, based on Github Icons

app/services/github/repositories/sync.rb
```diff
@@ -32,6 +27,33 @@ def call
...
+        repository_owner, repository_name = repository_owner_and_name(repository.link)
+        request = client.repos.get(repository_owner, repository_name)
+
+        if request.success?
+          repository.update!(
+            issues: request.body['open_issues_count'],
+            stars: request.body['stargazers_count']
+          )
+        end
+      end
```
tonyaraujop, line 56: can you organize the order of those methods based on the calling order? first is `update_repository_languages`, then `update_repository_stats` for instance. I think this might increase the readability a little bit. What do you think?

app/services/github/repositories/sync.rb
```diff
@@ -13,15 +13,10 @@ def call
         Rails.logger.info("[GH] -- Repositories: #{repositories.count}")
 
         repositories.map do |repository|
-          repository_owner, repository_name = repository.link.split('/')[-2..-1]
+          update_repository_languages(repository)
+          update_repository_stats(repository)
```
tonyaraujop, line 17: since we have two updates here, I think we can use `transaction` block. If everything should get updated if the request succeed right?

app/services/github/repositories/sync.rb
tonyaraujop, line 56:
```suggestion
      def update_repositoryrepository)
        repository_owner, repository_name = repository_owner_and_name(repository.link)
        request = client.repos.languages(repository_owner, repository_name)

        if request.success?
        languages = request.body.keys[0..MAX_LANGUAGES_NUMBER-1].join(', ')
        
          repository.update!(
            issues: request.body['open_issues_count'],
            stars: request.body['stargazers_count'],
            language: languages
          )
        end
      end
```

Since you have the same request with the same parameters and doing the same check, I think you can have just one single method `update_repository` and this would update all parameters at once. If you use this approach, I don't you need a `transaction` block btw. What do you think?

app/services/github/repositories/sync.rb
joaoGabriel55: I think that if one of them does not update the some of the infos: languages or stars and issues, this is not a problem, cause these infos are optional.

app/decorators/repository_decorator.rb
```diff
@@ -8,12 +8,30 @@ def self.collection_decorator_class
...
+
+  private
+
+  def format_number(number)
+    if number >= 1000
+      "#{(number/1000).round(1)}K"
+    else
+      number
+    end
+  end
```
fabiosammy, line 36: On rails we have a method called `number_to_human`:

https://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html#method-i-number_to_human

It helps to format this kind of approach.

Maybe something like:

```suggestion
  def format_number(number)
    number_to_human(number, precision: 1, units: {thousand: "k", billion: "kk"})
  end
```

It's also apply the i18n thing

db/migrate/20230728175727_add_issues_and_stars_to_repositories.rb
```diff
@@ -0,0 +1,6 @@
+class AddIssuesAndStarsToRepositories < ActiveRecord::Migration[7.0]
+  def change
+    add_column :repositories, :issues, :integer
+    add_column :repositories, :stars, :integer
```
fabiosammy, line 4: Can you add default values as `0` to those fields?

app/decorators/repository_decorator.rb
joaoGabriel55: Awesome! I did not know that, thanks for the tip!

app/decorators/repository_decorator.rb
```diff
@@ -1,19 +1,38 @@
...
+
+  def stars_formatted
+    format_number(model.stars)
+  end
+
+  private
+
+  def format_number(number)
+    return '0' if number.nil?
+    return number.to_s if number <= 999
```
abarbosa-godaddy, line 34: NIT: What if we set `0` as default?

```suggestion
  def format_number(number = 0)
    return number.to_s if number <= 999
```

app/services/github/repositories/sync.rb
```diff
@@ -13,15 +13,12 @@ def call
         Rails.logger.info("[GH] -- Repositories: #{repositories.count}")
 
         repositories.map do |repository|
-          repository_owner, repository_name = repository.link.split('/')[-2..-1]
+          languages = fetch_repository_languages(repository.link)
```
abarbosa-godaddy, line 16: I can see in [a comment below](https://github.com/Codeminer42/Punchclock/pull/479/files#diff-709bdfbc5d5e62f5efeb7dc03818d40bbab96c792086f334ee5a2917dff7f4b4R37) that this can be `nil`.

If `languages` is nil, should we really update the repository `language`?

app/decorators/repository_decorator.rb
joaoGabriel55: Nice tip! Thanks!

tonyaraujop (approved):
lgtm

app/decorators/repository_decorator.rb
```diff
@@ -1,19 +1,37 @@
...
+  end
+
+  def stars_formatted
+    format_number(model.stars)
+  end
+
+  private
+
+  def format_number(number)
+    return (number || 0).to_s if (number || 0) <= 999
```
abarbosa-godaddy, line 33: Honestly, It looks worse than before. Why not just let `0` be the default?

abarbosa-godaddy: you arr doing `(number || 0)` twice in a single line when setting `0` as default would solve it already.

app/services/github/repositories/sync.rb
abarbosa-godaddy: I don't get it. I didn't see any change or reply addressing this question :thinking: .

app/views/repositories/_repository.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      <span class="d-block text-dark"><%= simple_format repository.description %></span>
+    </div>
+    <div class='flex gap-4 items-center'>
+    <% if repository.issues.present? %>
+      <div class='flex gap-1 items-center text-gray-400' title="<%= "#{t("repositories.index.issues")}: #{repository.issues}" %>">
+        <%= inline_svg_tag 'issues.svg', class: 'h-4 w-4' %>
+        <span id='issues-tag' class='font-semibold'><%= repository.issues_formatted %></span>
+      </div>
+    <% end %>
+    <% if repository.stars.present? %>
```
adbatista, line 16: Why does this check when a repo will not have stars?

app/decorators/repository_decorator.rb
joaoGabriel55: I tested using default zero, but when the number is `nil` the default value is overrided.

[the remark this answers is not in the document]
joaoGabriel55: Let me improve

abarbosa-godaddy: I thought the `nil` was for when a value is not given. Why would it receive nil from somewhere?

(If a default value is not possible it's better to back how it was before)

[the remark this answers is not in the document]
abarbosa-godaddy: Question for you @adbatista . Should we be that defensive?

There's a check for `nil` in this method. Is it needed? In my humble opinion, If a client is passing `nil` for this method it's calling it incorrectly by definition.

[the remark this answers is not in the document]
adbatista: Yeah, I also don't think this method should have this concern

app/views/repositories/_repository.html.erb
joaoGabriel55: I decided to put these conditions for cases when for some reason a repository does not contains issues  count or stars count.

adbatista: Ok, I got the idea, but if we get at this point and the star, issues, etc are nil, we probably have a problem in some other part of the system, this guard might obfuscate the error

app/decorators/repository_decorator.rb
joaoGabriel55: I solved, just updating the `schema.rb` for `repositories` table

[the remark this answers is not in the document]
joaoGabriel55: I solved this concern about`nil`, just updating the `schema.rb`, adding default value as zero to issues and stars columns. I think this would be enough, what do you think?

app/decorators/repository_decorator.rb
```diff
@@ -1,19 +1,37 @@
...
+  end
+
+  def stars_formatted
+    format_number(model.stars)
+  end
+
+  private
+
+  def format_number(number)
+    return number.to_s if number <= 999
```
adbatista, line 33: why this? you are already formatting the number below

app/decorators/repository_decorator.rb
```diff
@@ -1,19 +1,37 @@
...
+
+    number_to_human(number, precision: 2, separator: '.', units: { thousand: "K", million: "M" }).gsub(/\s+/, '')
```
adbatista, line 35: why are you removing the spaces?

joaoGabriel55: Because, the output from `number_to_human` method is: `2 K` with space between the number and it unit.

app/decorators/repository_decorator.rb
joaoGabriel55: I'm doing this, 'cause I want to preserve the real count value from values bellow `999`, instead of use some kind of unit.

adbatista: what it is the return when you remove this and have a number less than 1k ?

joaoGabriel55: In the specific case of this test:
https://github.com/Codeminer42/Punchclock/blob/226d7faa9deb7843fb34b0ea00546941dd37f3ee/spec/decorators/repository_decorator_spec.rb#L39

If I remove that condition `number_to_human` method will return `1K` instead of `999`

app/decorators/repository_decorator.rb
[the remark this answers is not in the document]
adbatista: schema.rb is a auto generated file, yoou should not edit, change the migration to add a default value

app/decorators/repository_decorator.rb
adbatista: I see no problem in this case, but ok

app/decorators/repository_decorator.rb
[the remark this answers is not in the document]
joaoGabriel55: Sorry, I expressed myself wrong, I previously performed a rollback on the last migration I had created and created a new migration adding the new columns with default zero.

adbatista (approved):

PR #488, opened 2023-07-24
Order users' contributions by `created_at`

### LINKED ISSUE #487: Order the user contributions by the created instead of the link
In the "miners" view, reorder the contributions by the created_at date instead of the link. This will help the miner to update the right one when it's necessary:

https://punchclock.cm42.io/contributions

danielmbrasil:
Users' contributions are now ordered by newest ones first.

The contributions controller specs seemed to be mocking/stubbing the SUT so I changed that while I was in there. 

Closes #487.

### DIFF
--- a/app/controllers/contributions_controller.rb
+++ b/app/controllers/contributions_controller.rb
@@ -4,7 +4,7 @@ class ContributionsController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    @user_contributions = scopped_contributions.approved.order(pr_state: :desc)
+    @user_contributions = scopped_contributions.approved.order(created_at: :desc)
   end
 
   def edit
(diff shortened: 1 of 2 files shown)


app/controllers/contributions_controller.rb
```diff
@@ -4,7 +4,7 @@ class ContributionsController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    @user_contributions = scopped_contributions.approved.order(pr_state: :desc)
```
danielmbrasil, line 7: Should it be ordered only by `created_at` or also by `pr_state`?

fabiosammy: hmmm, good catch, I think that both, maybe the priority can be the pr_state and then the created_at. What do you think @adbatista ?

fabiosammy (approved):

app/controllers/contributions_controller.rb
fabiosammy: @danielmbrasil let's leave with the created_at only.

adbatista (approved):

Issue #502, opened 2023-07-27
Remove relation between offices and regional holidays


thiagofportella:
Since the table `offices_regional_holidays` was dropped in this PR: https://github.com/Codeminer42/Punchclock/pull/398, it is necessary to remove the association, in the model layer, to be able to destroy a Regional Holiday record with no errors raised.

PR #503, opened 2023-07-27
feat: Destroy action for regional holidays

thiagofportella:
Closes https://github.com/Codeminer42/Punchclock/issues/501 and https://github.com/Codeminer42/Punchclock/issues/502

### DIFF
--- a/app/models/regional_holiday.rb
+++ b/app/models/regional_holiday.rb
@@ -2,13 +2,9 @@
 
 class RegionalHoliday < ApplicationRecord
   has_and_belongs_to_many :cities
-  has_and_belongs_to_many :offices
+
   validates :name, :day, :month, presence: true
   validate :date_validation
-
-  def deprecated_cities
-    offices.map(&:city)
-  end
 
   def self.to_formatted_hash
     all.map { |h| { month: h.month, day: h.day } }
--- a/app/views/new_admin/regional_holidays/_regional_holidays_table.html.erb
+++ b/app/views/new_admin/regional_holidays/_regional_holidays_table.html.erb
@@ -10,7 +10,7 @@
   </thead>
   <tbody class="bg-white divide-y dark:divide-gray-700 dark:bg-gray-800 text-sm text-start">
     <% regional_holidays.each do |holiday| %>
-      <tr id=<%= "regional_holiday_#{holiday.id}" %> class="text-gray-700 dark:text-gray-400">
+      <tr id="<%= "regional_holiday_#{holiday.id}" %>" class="text-gray-700 dark:text-gray-400">
         <td class="px-4 py-3">
           <div>
             <%= holiday.name %>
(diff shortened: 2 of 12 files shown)


config/routes.rb
```diff
@@ -32,9 +32,10 @@
     resources :allocations, only: %i[show update edit], as: :user_allocations
     resources :users, only: %i[show update edit], as: :admin_user
     resources :punches, only: :show, as: :user_punches
-    resources :regional_holidays, except: %i[show update]
+    resources :regional_holidays, except: %i[show update destroy]
     resources :regional_holidays, only: :show, as: :show_regional_holiday
     resources :regional_holidays, only: :update, as: :update_regional_holiday
+    resources :regional_holidays, only: :destroy, as: :destroy_regional_holiday
```
adbatista, line 38: now that I noticed these duplicated declarations of routes, do you know why?

app/views/new_admin/regional_holidays/_regional_holidays_table.html.erb
```diff
@@ -10,7 +10,7 @@
   </thead>
   <tbody class="bg-white divide-y dark:divide-gray-700 dark:bg-gray-800 text-sm text-start">
     <% regional_holidays.each do |holiday| %>
-      <tr id=<%= "regional_holiday_#{holiday.id}" %> class="text-gray-700 dark:text-gray-400">
+      <tr id="<%= "regional_holiday_#{holiday.id}" %>" class="text-gray-700 dark:text-gray-400">
```
adbatista, line 13: you can use field_id here https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html#method-i-field_id

app/models/regional_holiday.rb
```diff
@@ -2,14 +2,10 @@
 
 class RegionalHoliday < ApplicationRecord
   has_and_belongs_to_many :cities
-  has_and_belongs_to_many :offices
```
adbatista, line 5: the migration, to drop the table is missing

thiagofportella: This migration was already merged in this PR: https://github.com/Codeminer42/Punchclock/pull/398/files#diff-7512307fbce008e0793ceff6f860d8dd6649aa2e7687be9d82e27261868bbe48

The removed code was meant to keep legacy registers working, but I reckon it's not needed anymore.

config/routes.rb
thiagofportella: I intended to name specific resources for `regional_holidays` in order to not conflict with ActiveAdmin's regional_holidays routes. In this case, `show`, `update` and `destroy` were declared separately to achieve this goal.

app/views/new_admin/regional_holidays/_regional_holidays_table.html.erb
thiagofportella: @adbatista, reading about this method, it seems that it's meant to be used in a ERB form helper, such as `text_field`. Any ideas how it would be assigned to a HTML `<tr>` tag?

adbatista (approved):

Issue #505, opened 2023-07-27
Fix flaky spec in vacation_mailer_spec


thiagofportella:
Log:
```
Failures:

  1) VacationMailer vacation email when a vacation is cancelled by the user renders the receivers emails
     Failure/Error: expect(mail.to).to eq(vacation_managers.map(&:email))

       expected: ["darrick@klein-wolf.org", "minda_hackett@ritchie-kessler.name"]
            got: ["minda_hackett@ritchie-kessler.name", "darrick@klein-wolf.org"]

       (compared using ==)
     # ./spec/mailers/vacation_mailer_spec.rb:165:in `block (4 levels) in <top (required)>'
```

PR #506, opened 2023-07-27
test: Fix flaky vacation mailer spec

thiagofportella:
Closes https://github.com/Codeminer42/Punchclock/issues/505


adbatista (approved):
The order of emails is supposed not to change, but it is ok. This does not change the behavior.

PR #513, opened 2023-08-01
Feat/create projects index new admin

thiagofportella:


app/models/concerns/filters/projects_filter.rb
```diff
@@ -0,0 +1,14 @@
+module Filters
+  module ProjectsFilter
```
adbatista, line 2: this is used in only one place, I don't see need for this

thiagofportella: Thanks for the input!

adbatista (approved):

Issue #518, opened 2023-08-03
Add allocation management and revenue forecast to new admin Project show page.


thiagofportella:
Original page powered by ActiveAdmin:
<img width="1440" alt="Captura de Tela 2023-08-03 às 12 06 55" src="https://github.com/Codeminer42/Punchclock/assets/52546576/7fe18bc8-0690-46f4-a608-8cd4958f8b5a">
<img width="1440" alt="Captura de Tela 2023-08-03 às 12 07 23" src="https://github.com/Codeminer42/Punchclock/assets/52546576/1f922061-5e04-4adb-814f-7a62eb1d9f2b">

PR #519, opened 2023-08-03
feat: add allocations and revenue forecast to projects page in new admin

thiagofportella:
Closes https://github.com/Codeminer42/Punchclock/issues/518

- Fix tabButtons bug in navbar.js, adding a optional chaining in the script.
- Add allocations and revenue forecast informations for a project in the New Admin show page.
- Create separate resource for allocate users in a project.

### Project show page:
Captura de Tela 2023-08-04 às 17 50 18
Captura de Tela 2023-08-04 às 17 50 37

### Allocate User in Project page:
Captura de Tela 2023-08-04 às 17 59 52
Captura de Tela 2023-08-04 às 17 59 38


app/assets/javascripts/new_admin/navbar.js
```diff
@@ -51,7 +51,7 @@ tabButtons.forEach((buttonElement) => {
     setActiveButton(buttonElement)
   });
 });
-tabButtons[0].click();
+tabButtons[0]?.click();
```
adbatista, line 54: when the tab buttons will be null?

app/views/new_admin/projects/show.html.erb
```diff
@@ -21,7 +21,29 @@
...
+      <div id="project_allocations" class="my-4">
+        <%= render 'project_allocations', allocations: @allocations %>
+      </div>
+    <% end %>
+
+    <% if @revenue_forecast.any? %>
+      <div id="project_revenue_forecast" class="my-4">
+        <%= render 'project_revenue_forecast', revenue_forecast: @revenue_forecast %>
+      </div>
+    <% end %>
```
adbatista, line 47: you could use a render collection in those two

adbatista, line 47: you could use a render collection in those two

app/assets/javascripts/new_admin/navbar.js
thiagofportella: It seems that `Cannot read properties of undefined` is raised when no `data-tab-id` is found in a NewAdmin page.
The following example is from the `show` Project page, when no revenue forecast or allocation is found for the specific project, meaning, in this scenario, no `data-tab-id` (or `data-tab-content`):
Captura de Tela 2023-08-07 às 10 47 32

PR #522, opened 2023-08-07
feat: Set tracking for approved contributions

thiagofportella:
Closes https://github.com/Codeminer42/Punchclock/issues/521

- After an admin user sets the status to `approved` for a contribution, it also changes the `tracking` attribute to true.


adbatista (approved):

app/services/contributions_approval_service.rb
```diff
@@ -0,0 +1,29 @@
...
+  private
+
+  attr_accessor :contribution, :reviewer_id
+
+  def update_contribution_review
+    contribution.update_reviewer reviewer_id
+  end
+
+  def set_tracking
+    contribution.update(tracking: true)
```
adbatista, line 27: we could change this to avoid hit database twice
maybe a `#assign_reviewer`

thiagofportella: Thanks for the input!

Issue #551, opened 2023-10-09
Create index page for questionnaires in new Admin 


nehamamandelbaum:
Original page powered by active admin: 

<img width="1670" alt="Captura de Tela 2023-10-09 às 17 48 52" src="https://github.com/Codeminer42/Punchclock/assets/75209689/d0d4ed17-ba68-47dc-a0cb-8add5d3bfed5">

PR #559, opened 2023-10-16
Index questionnaires new admin

nehamamandelbaum:
This PR Completes the index page of new_admin questionnaires, described in issue #551 

How the UI looks: 
Captura de Tela 2023-10-16 às 13 29 48

Captura de Tela 2023-10-16 às 13 30 02

### DIFF
--- a/app/models/questionnaire.rb
+++ b/app/models/questionnaire.rb
@@ -11,6 +11,7 @@ class Questionnaire < ApplicationRecord
   accepts_nested_attributes_for :questions, allow_destroy: true
 
   scope :active, -> { where(active: true) }
+  scope :by_title_like, ->(title) { where("questionnaires.title ILIKE ?", "%#{title}%") if title.present? }
 
   enumerize :kind,  in: { english: 0, performance: 1 },
                     scope: :shallow,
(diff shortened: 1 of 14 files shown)


app/models/questionnaire.rb
```diff
@@ -11,6 +11,11 @@ class Questionnaire < ApplicationRecord
   accepts_nested_attributes_for :questions, allow_destroy: true
 
   scope :active, -> { where(active: true) }
+  scope :by_title_like, ->(title) { where("questionnaires.title ILIKE ?", "%#{title}%") if title.present? }
+  scope :by_kind, ->(kind) { where(kind:) if kind.present? }
```
adbatista, line 15: There's no need for the if
kind is already a required argument

the same for the other scopes

app/queries/new_admin/questionnaires_query.rb
```diff
@@ -0,0 +1,28 @@
...
+    def initialize(filters)
+      self.filters = filters
+    end
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def call
+      retrieve_questionnaires
```
adbatista, line 12: why not calling the retrieve_questionnaires content here?

app/models/questionnaire.rb
nehamamandelbaum: If I remove the ifs, in the cases where the argument is nil, it will try to find some record that matches the nil argument. For example: when I do Questionnaire.by_kind(nil), it will try to find a questionnaire that has a kind = nil. 

What I want is that in the cases where I pass nil, it will simply not filter anything.

adbatista: I saw how I're calling, got it

adbatista (approved):

Issue #555, opened 2023-10-13
Create index page for repositories on new admin


jonathanrib3:
Create index page for /new_admin/repositories based on active admin admin/repositories page.

PR #560, opened 2023-10-17
feat: add new admin repositories index page

jonathanrib3:
Closes issue #555 

Adds index page on _/new_admin/repositories_ route

To make the languages filter work i used a query object that uses the scope `by_languages` on Repository that gets all repositories of certain languages, being that the available languages are fetched through the scope `languages` that returns repositories from distinct available languages (except empty languages). The 'search by name' filter uses the scope `by_repository_name_like` which gets all repositories that contains the string defined on the search through the patterns `%search%` and `%search`, since the name could be at the end of the repository link.

The buttons to create a new repository, show, edit and delete are not implemented yet.

Demonstration: 

demo-555

Multiple filters:

demo-555 pt2


spec/queries/new_admin/repository_query_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    subject { described_class.new(**filter_params).call }
+
+    it 'returns all repositories that matches the search' do
+      expect(subject).to eq(expected_attributes)
+    end
+  end
+
+  context 'when filter options are blank' do
+    let!(:repositories_list) { create_list(:repository, 2) }
+    let(:filter_params) {
```
adbatista, line 66: use the do/end syntax here

app/queries/new_admin/repository_query.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    private
+
+    def languages
+      if @languages.nil?
+        return nil
+      end
+
+      @languages.reject(&:blank?)
+    end
```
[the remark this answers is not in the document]
jonathanrib3: When the page is loaded without any params, `languages` equals to `nil`, but when the page is loaded with a filter applied and no languages are selected, languages equals to `[""]`, which is why in this case i'm using this method, to return an empty array which the presence's going to be checked on `by_languages` scope. Is there a better to do this? That nil condition could be refactored but i don't know about the method itself...

app/models/ability.rb
```diff
@@ -28,6 +28,7 @@ def initialize(user)
 
     if user.admin? || user.open_source_manager?
       can :read, ActiveAdmin
+      can :manage, Repository
```
adbatista, line 31: add this in the admin_ability

jonathanrib3:
Refactored on e378c30bf437da30ea6f45ca2d6126e9684fe7ac

app/models/repository.rb
```diff
@@ -6,4 +6,12 @@ class Repository < ApplicationRecord
   def to_s
     link
   end
+
+  scope :languages, -> { select(:language).distinct.where.not(language: [nil, '']) }
```
adbatista, line 10: Why filter only to the language here?

if you want repositories with languages what about `Repository.with_languages`

app/models/repository.rb
```diff
@@ -6,4 +6,13 @@ class Repository < ApplicationRecord
...
+  scope :by_repository_name_like, lambda { |repository_name|
```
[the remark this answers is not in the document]
adbatista: you did not change to stabby lambda

app/models/repository.rb
```diff
@@ -6,4 +6,12 @@ class Repository < ApplicationRecord
...
+    if repository_name.present?
```
adbatista, line 12: what is the need of this check?

app/models/repository.rb
```diff
@@ -6,4 +6,13 @@ class Repository < ApplicationRecord
...
+      where("link ILIKE ?", "%#{repository_name}%")
+        .or(where("link ILIKE ?", "%#{repository_name}"))
```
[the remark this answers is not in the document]
adbatista: did you check this?

app/models/repository.rb
jonathanrib3: Some repositories has blank values (specially repositories from the seeds), it didn't make sense to me to show a blank language on the languages filter so i did this check

app/models/repository.rb
jonathanrib3: That scope is being used on repositories languages filter, which shows the available distinct languages from the repositories. Indeed that scope's name's not clear enough as it should be, on my way to change that

app/models/ability_admin.rb
```diff
@@ -43,13 +43,14 @@ def define_permissions_for(user)
   def admin_permitions(user)
     can :manage, action + [
       Punch
+
```
EduardoSCosta, line 46: remove this empty line :nail_care:

app/views/new_admin/repositories/_repositories_table.html.erb
```diff
@@ -0,0 +1,42 @@
...
+							class="flex items-center justify-between px-2 py-2 text-sm font-medium leading-5 rounded-lg text-primary-600 dark:text-gray-400 hover:text-primary-400 hover:dark:text-gray-300 focus:outline-none focus:ring-gray"
+							aria-label="Delete">
+          		<%= heroicon 'trash', variant: :mini, options: { class: "w-5 h-5" } %>
+						</button>
+					</div>
+				</td>
+			</tr>
+			<% end %>
+	</tbody>
+</table>
```
EduardoSCosta, line 42: add empty line at the end of the file

app/views/new_admin/repositories/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+			<%= render 'repositories_filters' %>
+		</div>
+	</div>
+</div>
+
+<% if @repositories.any? %>
+  <section id="pagination_regional_holidays" class="dark:text-gray-400">
+    <%= paginate @repositories %>
+  </section>
+<% end %>
```
EduardoSCosta, line 34: empty line here too

jonathanrib3:
Added EOF and removed empty line on 2a5135f91aeb5dde28119fdc3349a39feabd390e

app/models/repository.rb
adbatista: the argument is required, if the dev is passing an empty string it's the dev's fault

app/models/repository.rb
adbatista: I got it; my question was about the `select(:language)` you just want the the laguages,

so there are two problems here

* Scope is not the correct approach; the scope is used to filter (scope) the Repositories, and here you want an array of languages
* Te second problem is that the language column, can have multiple languages, it can have "Ruby", "Javascript" and can also have "Ruby, Javascrit", "Javascript, Ruby", "Ruby, React, javascript", etc ...

PR #578, opened 2023-10-20
Feat/education experience index

sergioosouzaa:
This pull request addresses issue https://github.com/Codeminer42/Punchclock/issues/561, which requires allowing users to view and edit their own resume-related information. I have implemented the index functionality of the educational experiences in this PR, and I will complete the CRUD operations in subsequent follow-up PRs.

User with education experiences:
user_with_experience

User without education experiences:
user_without_experience


app/controllers/education_experiences_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class EducationExperiencesController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
+    @undecorated_experiences = EducationExperience.for_user(current_user.id)
```
adbatista, line 7: why is this in an instance variable?

app/decorators/education_experience_pagination_decorator.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class EducationExperiencePaginationDecorator < Draper::Decorator
+  delegate_all
+
+  def initialize(params, object, options = {})
+    @params = params
+    @unpaginated_object = object
```
adbatista, line 8: is this variable necessary?

spec/features/navigation_bar_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    context 'when the user is logged in' do
+      before do
+        sign_in user
+      end
+
+      it 'has the education experience dropdown option on the navigation bar', agreggate_failures: true do
+        visit root_path
+
+        expect(page).to have_selector('.dropdown-toggle', text: t('refills.navigation.my_resume'))
+        click_link t('refills.navigation.my_resume')
```
adbatista, line 18: use the raw text instead of the translation helper, this way if there's a problem with the translation, the task will catch

spec/features/navigation_bar_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+feature 'Navigation Bar' do
+  describe 'Resume options' do
+    let(:user) { create(:user) }
+
+    context 'when the user is logged in' do
```
adbatista, line 9: The test for unlogged users is missing

spec/models/education_experience_spec.rb
```diff
@@ -38,4 +38,18 @@
       it { expect(education_experience).to_not be_valid }
     end
   end
+
+  describe '.for_user' do
+    let!(:user) { create(:user) }
+    let!(:education_experience_from_user) { create(:education_experience, user:) }
+    let!(:education_experience_from_another_user) { create(:education_experience) }
```
adbatista, line 45: the `let!` is not necessary here

spec/models/education_experience_spec.rb
```diff
@@ -38,4 +38,18 @@
...
+
+    it 'returns education experiences associated with the specified user' do
+      expect(EducationExperience.for_user(user.id)).to include(education_experience_from_user)
```
adbatista, line 48: You should match the correct response, only checking if it includes the experience, does not verify the correct return, you should verify the full return

app/decorators/education_experience_pagination_decorator.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class EducationExperiencePaginationDecorator < Draper::Decorator
```
adbatista, line 3: why are you using this approach instead of using a collection decorator?

adbatista (changes requested):

PR #581, opened 2023-10-20
feat: add hourly rates sorts and filters on /admin/contributions

### LINKED ISSUE #576: Ajustes no index das alocações no ActiveAdmin
Precisamos fazer alguns ajustes na tabela da página de index das alocações do ActiveAdmin.
* Permitir que o usuário possa ordenar pela taxa horária
* Adicionar um filtro onde será possível selecionar alocações que são cobradas em dólar ou real

Link para a página no staging: http://punchclock-staging.herokuapp.com/admin/allocations

jonathanrib3:
### Problem

Closes #576 - Adjustments on ActiveAdmin Allocations' index page

### Solution

To make the hourly rates sorting work i decided to sort allocations through their value as cents - `hourly_rate_cents` and used ActiveAdmin's `sortable` property on the table structure. 

To make the hourly rates currencies filter work i added a scope that gets all available currencies from allocations table without any duplicates in ascending order, being that they're rendered through an ActiveAdmin `filter` as a select. 

A demo of how things are working right now:

demo-576

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -23,9 +23,7 @@ class Allocation < ApplicationRecord
   scope :ongoing, -> {
     where(ongoing: true, user_id: User.active).order(start_at: :desc)
   }
-  scope :hourly_rate_currencies, -> {
-    select(:hourly_rate_currency).distinct.map(&:hourly_rate_currency)
-  }
+
   scope :finished, -> {
     where(ongoing: false, user_id: User.active).order(end_at: :desc) }
   scope :in_period, -> (start_at, end_at) do
@@ -46,6 +44,10 @@ class Allocation < ApplicationRecord
     else
       I18n.t('finished')
     end
+  end
+
+  def self.hourly_rate_currencies
+    select(:hourly_rate_currency).distinct.pluck(:hourly_rate_currency)
   end
 
   def user_punches
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -23,6 +23,9 @@ class Allocation < ApplicationRecord
   scope :ongoing, -> {
     where(ongoing: true, user_id: User.active).order(start_at: :desc)
   }
+  scope :hourly_rate_currencies, -> {
+    select(:hourly_rate_currency).distinct.map(&:hourly_rate_currency)
+  }
   scope :finished, -> {
     where(ongoing: false, user_id: User.active).order(end_at: :desc) }
   scope :in_period, -> (start_at, end_at) do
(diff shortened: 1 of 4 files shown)


spec/models/allocation_spec.rb
```diff
@@ -128,4 +128,21 @@
...
     end
   end
+
+  describe '#hourly_rate_currencies' do
+    let!(:brl_allocations) { create_list(:allocation, 2, hourly_rate_currency: 'BRL') }
+    let!(:usd_allocation) { create(:allocation, hourly_rate_currency: 'USD') }
+
+    subject { described_class.hourly_rate_currencies }
+
+
```
carloslopes, line 138: Unnecessary newline 💅

spec/models/allocation_spec.rb
```diff
@@ -128,4 +128,21 @@
...
+    it 'gets the available currencies from allocations without duplicates in ASC order' do
+      expect(subject).to eq (
+        [ 
+          usd_allocation.hourly_rate_currency,
```
carloslopes, line 142: It's always good to avoid as much as possible any logic on expectations. In this case, since you know exactly what you are expecting based on your setup, you could do this way: `expect(subject).to eq(['USD', 'BRL'])`

spec/models/allocation_spec.rb
jonathanrib3: When it comes to specs/examples structure i get confused sometimes, isn't it better to explicit the subject on each context/describe? By doing this it gets more confusing for those reading the test? Or does it gets easier?

carloslopes (approved):

app/models/allocation.rb
```diff
@@ -23,6 +23,9 @@ class Allocation < ApplicationRecord
   scope :ongoing, -> {
     where(ongoing: true, user_id: User.active).order(start_at: :desc)
   }
+  scope :hourly_rate_currencies, -> {
+    select(:hourly_rate_currency).distinct.map(&:hourly_rate_currency)
+  }
```
adbatista, line 28: This should be a class method since it does not return a relation; in this case, it would be better to use pluck. This way, it does not allocate unnecessary memory.

jonathanrib3: Refactored on f6ff0c5e9dd160c195eaffea20da6b0fc796afd

app/models/allocation.rb
```diff
@@ -54,6 +59,8 @@ def user_punches
       .decorate
   end
 
+
+
```
adbatista, line 63: extra space

jonathanrib3: Removed on d610d771633ab737a460bfb7468c0a604431272b, thanks for the catch

adbatista (approved):

PR #583, opened 2023-10-23
feat: add hourly rates to allocations table on /admin/users page

jonathanrib3:
### Problem

**Closes issue**  #577 - Add hourly rate column to allocations table

### Solution

To solve this, I simply added on admin User's code base a column named `hourly rate`. The values in that column are going to be rendered using Allocation's Decorator, which already has a function that formats the hourly rate with their right currency symbol and amount.

Here's the result:

image

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -130,6 +130,9 @@ ActiveAdmin.register User do
               column :project_name do |allocation|
                 allocation.project.name
               end
+              column :hourly_rate do |allocation|
+                allocation.decorate.hourly_rate
+              end
               column :start_at
               column :end_at
               column :ongoing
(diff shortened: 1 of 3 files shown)


carloslopes (approved):

app/admin/user.rb
```diff
@@ -130,6 +130,9 @@
               column :project_name do |allocation|
                 allocation.project.name
               end
+              column :hourly_rate do |allocation|
+                allocation.decorate.hourly_rate
```
adbatista, line 134: Use the helper of active admin to decorate the resource instead of calling here, if you have another decorated method, this would be called every time

jonathanrib3: Refactored on d854b1634a668afc8461d214596a7f99ca3eef46

PR #606, opened 2023-11-08
Refactor new admin

EduardoSCosta:
- [x] Add new admin controller;
- [x] Add authorization to controllers;
- [x] Fix broken specs.

### DIFF
--- a/app/controllers/new_admin_controller.rb
+++ b/app/controllers/new_admin_controller.rb
@@ -1,0 +1,11 @@
+class NewAdminController < ApplicationController
+  include NewAdmin::Pagination
+
+  before_action :authenticate_user!
+
+  private
+
+  def current_ability
+    @current_ability ||= AbilityAdmin.new(current_user)
+  end
+end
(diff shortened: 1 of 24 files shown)


nehamamandelbaum (approved):

app/controllers/new_admin_controller.rb
```diff
@@ -0,0 +1,11 @@
+class NewAdminController < ApplicationController
+  include NewAdmin::Pagination
+
+  before_action :authenticate_user!
```
adbatista, line 4: your are missing the `load_and_authorize_resource` here

EduardoSCosta: If I put `load_and_authorize_resource` in `NewAdminController`, controllers without a model will break:

For example:
image

adbatista: So this needs to be explicitly added for all new admin controllers

adbatista (approved):

PR #650, opened 2023-12-11
523: Fixes query for data-tab-id elements in new admin

### LINKED ISSUE #523: Fix query for `data-tab-id` elements in `new_admin/navbar.js`
It seems that the script is querying for those elements in all New Admin pages.
This PR https://github.com/Codeminer42/Punchclock/pull/519 adds a null safety check, however a more in-depth investigation may be necessary.

ghost:
Fix #523 
I propose removing the `data-tab-id` attributes from new admin divs to prevent confusion with the attributes applied to new admin buttons, which are the target elements in navbar.js.
The div attributes do not appear to be used anywhere.

### DIFF
--- a/app/assets/javascripts/new_admin/navbar.js
+++ b/app/assets/javascripts/new_admin/navbar.js
@@ -1,4 +1,4 @@
-const tabButtons = Array.from(document.querySelectorAll("[data-tab-id]"));
+const tabButtons = Array.from(document.querySelectorAll("button[data-tab-id]"));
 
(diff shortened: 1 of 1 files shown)


thiagofportella (commented):
Seems to be working fine. Good job.

It would be nice to test the behaviour of `data-tab-id` buttons by writing some capybara specs for pages that have those buttons (if there are no such specs already).

Also, it's a good idea to remove those data attributes from the divs in this PR.

